PR #3, opened 2015-02-04
Redirect system email to web-administrators list

henare:
What do you think @mlandauer?


mlandauer:
Looks good. Do we know that there is an MTA installed at this stage?

if not we should add that as a dependency

henare:
>Looks good. Do we know that there is an MTA installed at this stage?

Yeah Postfix is installed by default so I think it's fine.

henare:
:tanabata_tree: thanks @mlandauer :tanabata_tree:

Issue #8, opened 2015-11-14
run-with-lockfile.sh is missing, what am I doing wrong?


andreicristianpetcu:
Hi, 

For some reason [this task](https://github.com/andreicristianpetcu/infrastructure/blob/nvs/roles/righttoknow/tasks/main.yml#L42-L46) fails. I cloned commonlib and ansible passed but what is the correct way to do it? I disabled non-alaveteli sites so this might be the cause.

``` bash
#mkdir -p /srv/www/current/
#cd /srv/www/current/
#git clone https://github.com/mysociety/commonlib.git
```

I ran using [my fork](https://github.com/andreicristianpetcu/infrastructure) of infrastructure, the secret is.... well.... `secret` in ~/.infrastructure_ansible_vault_pass.txt :)

andreicristianpetcu:
I made a small [hack](https://github.com/andreicristianpetcu/infrastructure/commit/704ea6b471ef2d3e6e2de7a1e4c4521b772e8954) in my fork. Not sure if this is the correct way.

henare:
>I guess ansible is used only for provisioning, not for deployment.

[That's correct](https://github.com/openaustralia/infrastructure#right-to-know).

However I think you may have found an actual issue here. [This file](https://github.com/andreicristianpetcu/infrastructure/blob/nvs/roles/righttoknow/tasks/main.yml#L50) isn't going to exist until you deploy.

andreicristianpetcu:
@henare I made a [pull request](https://github.com/openaustralia/infrastructure/pull/9) that adds the missing run-with-lockfile.sh file.

PR #9, opened 2015-11-15
#8: Make user mysociety commonlib is present

andreicristianpetcu:


henare:
This gets deployed by Capistrano so it doesn't feel right to have both Ansible and Capistrano managing it.

I'm not sure what the right solution is though :smiley:

andreicristianpetcu:
> I'm not sure what the right solution is though :smiley: 

Well.... that's why I asked. Is the Capistrano deploy idempotent? (I am working on using it right now and I did not manage yet to deploy it). I mean, if both Ansible and Capistrano check if commonlib exists and if it, they skip it then it should be fine.

The problem I was trying to fix is that ansible shoud provision the server, but it expects that a capistrano deploy was done and so ansible fails. Capistrano expects that ansible provisioned the server with no errors.

I will look into this problem after I manage to finish the capistrano deploy.
I will do all the necessary dirty hacks in order to finish the setup and then try to fix the problems in the correct way and push them upstream.

andreicristianpetcu:
This is not a good pull request :) The issue is there, both Ansible and Capistrano deploy commonlib but we need to find a better solution. Can you cancel this pull request? :)

PR #44, opened 2018-03-04
WIP: Update electionleaflets

jamezpolley:
* So far: use modern mysql config
* Next: update package list - "php5" is failing to install


jamezpolley:
I'm working on the assumption that we'd like to change as little as possible to get this running. So I'm not planning to change apache2 for nginx; I'm not planning to change any of the cap tasks unless they are obviously broken; I'm not planning to fix things like the deprecated use of sudo. At least, not in this pass, they might get tackled seperately, later or in parallel.

This seems to get the VM up enough for capistrano to attempt a deployment.. which fails, but that means I can move on to digging into the capistrano tasks in the electionleaflets repo

mlandauer:
@jamezpolley sounds like you made some really solid progress there and everything you say makes perfect sense. 💯

PR #76, opened 2018-03-28
Update hostnames for electionleaflets prod to point at ec2

jamezpolley:


ec2-hosts
```diff
@@ -6,7 +6,7 @@ righttoknow.org.au
 openaustraliafoundation.org.au
 opengovernment.org.au
 ec2.opengovernment.org.au
-ec2.electionleaflets.org.au
+electionleaflets.org.au
```
mlandauer: Don't need to change this one until after the migration is done

terraform/electionleaflets-dns.tf
```diff
@@ -3,39 +3,11 @@ resource "cloudflare_record" "el_root" {
   domain = "electionleaflets.org.au"
   name   = "electionleaflets.org.au"
   type   = "A"
-  value  = "${aws_eip.kedumba.public_ip}"
-}
-
-resource "cloudflare_record" "el_ec2" {
```
mlandauer: I wouldn't remove the ec2 records until after the migration is done. Then you don't have to change so much in one go.

PR #78, opened 2018-03-29
Add php newrelic agent

### LINKED ISSUE #77: Electionleaflets app is not reporting in newrelic
We're getting the info from the server but not the application

jamezpolley:
Fixes #77


jamezpolley:
Have confirmed that data from app is visible in newrelic

PR #99, opened 2018-10-03
Move towards dev deployment without needing a password

jamezpolley:
- Adds a hostvars directory which can provide unencrypted values
   specifically for the vagrant hosts
 - Remove the default password file from ansible.cfg
   With this value set, Ansible will attempt to read the pasword even if
   it's not used.
 - Modify comments in site.yml to show use of --vault-id option
 - Modify .gitignore to ignore the .vault_pass.txt file


jamezpolley:
I'm going to close this in favour of Ben's gnuplot branch which extends this

PR #98, opened 2018-10-03
Add SPF/DKIM records for Front

benrfairless:
Resolves #97

PR #101, opened 2018-10-04
Update Morph.io DNS for Front

benrfairless:
I missed something in #98 so this is the update. I'm going to merge this immediately.


benrfairless:
@jamezpolley when you get a min can you re-deploy Morph to pick up the change :) Thanks!

PR #103, opened 2018-10-17
Add client_max_body_size 100M

benrfairless:
Attempts to resolve https://github.com/openaustralia/righttoknow/issues/705 by updating nginx.conf and both staging and default files to 100MB files.

To confirm this is working, once deployed go to https://www.test.righttoknow.org.au/upload/request/testing_uploads and attempt to upload a 10MB file (should work) and a 101MB file (should fail).


jamezpolley:
This time, I was able to use the test instance to confirm tha this works.

Thanks Ben

PR #104, opened 2018-10-18
Secretless Dev + GNUplot

jamezpolley:
Progress towards being able to set up a dev environment without needing ansible secrets; and to using different secrets for each environment.

Also adding in gnuplot for graphing. Because that's a thing that happened at the same time.


jamezpolley:
Ben and I have both been using this for local dev. I'm going to go ahead and merge it so that we can move forward on the passwordless-dev project, and also towards getting Ben access to the RTK machine

PR #119, opened 2019-01-08
Add new API key for use on website

jamezpolley:
This adds the new key required for use with
openaustralia/planningalerts/pull/1327

The existing key is renamed to be the "email" key, as we can't change
the key that's been used on emails that have already been sent out.

The existing key name is applied to the new key, so that new web
traffic will start using the new key immediately.

Connects to openaustralia/planningalerts/issues/1323


mlandauer:
I'm going to merge this now and run the provisioning so things are ready to merge the PR on PlanningAlerts itself. The only thing I changed was the referrers on the new api key to be exactly the same as the one that currently works in production. I don't I'm willing right now to risk breaking something here.

mlandauer:
Now deploying this to production...

mlandauer:
Done.

PR #164, opened 2022-03-07
Add Txt Record for Microsoft Domain Verification

benrfairless:
Kat agreed to register us for Microsoft 365 for NonProfits. This is my attempt to create a TXT record on oaf.org.au with the below variables:

image


mlandauer:
Sorry I don't understand @benrfairless. Why is this needed? What's the end goal?

katska:
@benrfairless are you able to use the tools you need to without hooking it up this way? We're not considering switching over to use it for everyone.

PR #168, opened 2024-09-02
Change DMARC Policy on Right to Know to quarantine

benrfairless:
I've confirmed that emails are meeting the required policy. Can we please increase the DMARC level from just reporting to quarantine now?


simonzippy (approved):
This DMARC record is valid so good to go. @benrfairless - once this has updated the DNS on CloudFlare we should check an email sent from righttoknow to a gmail account as we can monitor how DMARC is processed easily in gmail by looking at the headers as you likely already know. Also - do we know from the postmarkapp.com service if anyone is or has been trying to spoof us?

PR #171, opened 2025-03-06
Block aggressive bot from crawling RTK

jamezpolley:
List of IPs copied from https://github.com/openaustralia/infrastructure/commit/3279f93379d6df38029334dc19ca5213ace49fa4 as the same IPs now seem to be hitting RTK.

### DIFF
--- a/roles/internal/righttoknow/templates/nginx/stage
+++ b/roles/internal/righttoknow/templates/nginx/stage
@@ -74,6 +74,14 @@ server {
   {% endif %}
 
   location / {
+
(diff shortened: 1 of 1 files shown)


katska:
@mlandauer Any reason not to merge this? Sorry to poke in here maybe wrongheadedly - review / merge? Either way please take a look and comment if it's not the write thing.

mlandauer:
@jamezpolley I'll merge this now (I don't know why I didn't get round to merging it before). Note however that the block list for TVFY got even bigger. So, it might be worth extending it similarly if there's more problems with RTK?

PR #172, opened 2025-05-09
Add missing symlink for  .vault_pass.txt

jamezpolley:
Other vault passphrase files had symlinks, but this one was missing its symlink.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 
 .vagrant
-.vault_pass.txt
 roles/external
 test.pem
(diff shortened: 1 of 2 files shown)


jamezpolley:
@simonzippy This should get you able to run vagrant (once Keybase sync the new file)

jamezpolley:
I want to double-check that I've got the right file before committing this.

@mlandauer could you check that the contents of `/keybase/team/oaforgau.sysadmin/.vault_pass.txt` match your `.vault_pass.txt`?

PR #174, opened 2025-05-09
Remove unused plausible role

jamezpolley:
According to comments inside the files, it seems that the plausible role was intended for building a plausible host - but we've decided to go with the hosted service for the time being.

Having the role included has lead to problems for me trying to build dev VMs. If I'm understanding what's happening, the Vagrantfile doesn't define a host group called `plausible`; and this leads to Ansible trying to apply the role on all hosts. The cleanest solution to this seems to be to remove the unused role and leave only code that is being used.

In addition, the Plausible role requires ansible 2.10; comments in the files indicate that we don't know if this works for any other role. Again, the cleanest solution seems to be to remove the unused role so that we have a single ansible version we know works for all current roles.


mlandauer:
LGTM. Thanks @jamezpolley

PR #182, opened 2025-05-30
Add missing dependency + docs re packer build

jamezpolley:
- Amazon plugin is required to use the `amazon-ebs` source
- Adding notes to docs to show where to install packer, and a brief mention in the main README

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,18 @@ bundle exec cap development deploy
 
 #### Deploying PlanningAlerts to production
 
+We now have two productions servers, and a blue/green deployment process driven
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -349,6 +349,19 @@ bundle exec cap development deploy
 
 #### Deploying They Vote For You to production
 
+We now have two productions servers, and a blue/green deployment process driven
```
mlandauer, line 352: Oops. Looks like this got added to the They Vote For You section rather than the Planning Alerts section

mlandauer (changes requested):
Looks good to me other than the addition to the main README I think is in the wrong section

PR #196, opened 2025-08-02
Removing Twitter from RightToKnow

Br3nda:


Br3nda:
Note: this is already applied to production

simonzippy (approved):
All good and already in Prod! :D

PR #202, opened 2025-08-27
adding tags to user key task

Br3nda:
So it can be applied on it's own without running all the other tasks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.phony: venv roles production ALL letsencrypt
+.phony: venv roles production ALL letsencrypt check-rtk
 
 ALL: venv roles .vagrant
 
@@ -40,3 +40,11 @@ clean:
 	
 clean-all: clean
 	rm -rf .vagrant
+
+check-righttoknow:
+	.venv/bin/ansible-playbook -i ./inventory/ec2-hosts site.yml -l righttoknow --check
+check-planningalerts:
+	.venv/bin/ansible-playbook site.yml -l planningalerts --check
+
+apply-keys:
+	.venv/bin/ansible-playbook site.yml --tags userkeys
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.phony: venv roles production ALL letsencrypt check-rtk
+.PHONY: venv roles production ALL letsencrypt check-rtk
 
 ALL: venv roles .vagrant
 
--- a/Makefile
+++ b/Makefile
@@ -46,5 +46,5 @@ check-righttoknow:
 check-planningalerts:
 	.venv/bin/ansible-playbook site.yml -l planningalerts --check
 
-apply-keys:
+update-github-ssh-keys:
 	.venv/bin/ansible-playbook site.yml --tags userkeys
--- a/roles/internal/righttoknow/tasks/main.yml
+++ b/roles/internal/righttoknow/tasks/main.yml
@@ -353,7 +353,7 @@
     - staging
   notify: nginx reload
 
-- name: Copy across the main Alaveteli config (production)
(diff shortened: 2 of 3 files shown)


roles/internal/righttoknow/tasks/main.yml
```diff
@@ -353,7 +353,7 @@
     - staging
   notify: nginx reload
 
-- name: Copy across the main Alaveteli config (production)
+- name: Apply Alaveteli config (production)
```
benrfairless, line 356: Is there any particular reason we are renaming these tasks? It's not clear the benefit of doing this.

roles/internal/righttoknow/tasks/main.yml
```diff
@@ -374,7 +374,7 @@
     pro_referral_coupon: "{{ pro_referral_coupon_production }}"
   notify: nginx restart
 
-- name: Copy across the main Alaveteli config (staging)
+- name: Apply Alaveteli config (staging)
```
benrfairless, line 377: Same as my [earlier comment](https://github.com/openaustralia/infrastructure/pull/202#discussion_r2302500038).

roles/internal/righttoknow/tasks/main.yml
```diff
@@ -440,7 +440,7 @@
 # Generated public/private keypair for opendkim with:
 # opendkim-genkey -b 2048 -d righttoknow.org.au -D /etc/dkimkeys -s default -v
 
-- name: Copy across DKIM keypair
+- name: Apply DKIM keypair
```
benrfairless, line 443: Same as my [earlier comment](https://github.com/openaustralia/infrastructure/pull/202#discussion_r2302500038).

benrfairless (approved):

Makefile
```diff
@@ -39,4 +39,12 @@ clean:
 	rm -rf .venv roles/external site.retry collections 
 	
 clean-all: clean
-	rm -rf .vagrant
\ No newline at end of file
+	rm -rf .vagrant
+
+check-righttoknow:
```
jamezpolley, line 44: This seems unrelated to the change described in the title. Since this repo is fairly new to being managed this way, I'm not strongly objecting to letting this through anyway, but I wanted to note that my general preference would be for unrelated changes to be in their own PRs

(It should also be noted that most of my experiences has involved tools like Google's Mondrian, or the open source clone Gerrit (https://www.gerritcodereview.com/), which both have tools that make it much easier to handle chains of pull requests than stock Git or Github PRs - so it's possible my preferences are not well aligned with the tooling in this case)

jamezpolley (approved):
There's a separate playbook which already exists in order to apply user keys to machines without running other tasks - https://github.com/openaustralia/infrastructure/blob/master/update_ubuntu_keys.yml

I do think this approach is better though as it avoids the possibility of what's in that file drifting away from what's in the main playbooks. 

I can't find any references to this in the documentation, so I don't think there's any documentation to update to remove references to the old procedure. 

Is it worth removing update_ubuntu_keys.yml? Or maybe replacing it with a pointer to the Makefile?

Is it worth adding mention of this to the documentation? My gut feeling is that the reference in the Makefile is sufficient.

Overall, there are a few things in this that I have a weak preference for being done differently, but they're small things and outweighed by the benefits of this change. I don't have any objection to this landing as-is

roles/internal/righttoknow/tasks/main.yml
Br3nda: It sounds to me like a much clearer description of what it's doing here.

Makefile
Br3nda: i think there are "Release Chain" like tools in github for this. But agreed, separate things in separate PRs is better

PR #206, opened 2025-08-31
Feature/planning alerts inventory

Br3nda:
Adding planning alerts host names into inventory


benrfairless (commented):
Probably not my place to review this. Would suggest @jamezpolley have a look.

simonzippy (approved):
- We know the updated ansible playbook command reference works from tests on other services 
- and I've checked the domain/DNS for the 2 PlanningAlerts EC2 instances are correct

...so this is good to go.

PR #208, opened 2025-09-04
Update morph ssl certificate and clarify where generate-certificates.sh is run from

ianheggie-oaf:
Ensure generate-certificates.sh is run from the project root and moves across morph cert if morph repo is present

### DIFF
--- a/certificates/generate-certificates.sh
+++ b/certificates/generate-certificates.sh
@@ -3,6 +3,13 @@
 # Generates certificates for local development. A single certificate, that
 # is unique to you, is added to your browser
 
+if ! [ -f certificates/generate-certificates.sh ]; then
+  echo ERROR: Must be run in project root directory!
+  exit 1
+fi
+
+cd certificates || exec echo Failed to cd to certificates directory!
+
 # Note that we're not generating certificates for PlanningAlerts because those
 # are served from the load balancer on AWS so the server itself only needs
 # to support http and so doesn't need a certificate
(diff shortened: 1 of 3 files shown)


certificates/generate-certificates.sh
```diff
@@ -3,6 +3,13 @@
 # Generates certificates for local development. A single certificate, that
 # is unique to you, is added to your browser
```
Br3nda, line 5: would it be a good idea to do this near the top?

```suggestion
set -eu
```

certificates/generate-certificates.sh
```diff
@@ -57,14 +64,29 @@ do
...
+# morph provisioning is moved to this repo
+
+if [ -d ../../morph/provisioning/roles/morph-app ]; then
+  echo Moving local ssl certificate to morph ...
+  set -x
+  mkdir -p ../../morph/provisioning/roles/morph-app/files/ssl
+  mv dev.morph.io.key dev.morph.io.pem ../../morph/provisioning/roles/morph-app/files/ssl
+  set +x
+else
+  echo Missing ../../morph/provisioning/roles/morph-app directory: skipped copying dev.morph.io files ...
```
Br3nda, line 90: do we want to `exit 1` here maybe?

certificates/generate-certificates.sh
```diff
@@ -57,14 +64,29 @@ do
...
+# FIXME: adjust temporary measure to copy across a certificate generated here
+# to be used in the morph repo for provisioning once
+# morph provisioning is moved to this repo
+
+if [ -d ../../morph/provisioning/roles/morph-app ]; then
+  echo Moving local ssl certificate to morph ...
+  set -x
+  mkdir -p ../../morph/provisioning/roles/morph-app/files/ssl
+  mv dev.morph.io.key dev.morph.io.pem ../../morph/provisioning/roles/morph-app/files/ssl
+  set +x
```
Br3nda, line 88: so, why? Is it trying to say it's fine for this to fail?

Br3nda: oh.. i see. it's just trying to be quieter here.

Br3nda:
these all look like great improvements to me. I'll look for some time to test run it myself soon

certificates/generate-certificates.sh
ianheggie-oaf: Added pipefail and reporting as well

ianheggie-oaf:
@Br3nda please check rebased and updated commits - it should be the best of both our input

Br3nda (approved):

PR #210, opened 2025-09-04
Add convienience ansible variables

ianheggie-oaf:
i added the same two convenience vars I added in morph for consistency - speeds up playbook development


Vagrantfile
```diff
@@ -89,6 +89,15 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         "redis.test"
       ]
     }
+    raw_args = []
+    if ENV['TAGS']
+      raw_args <<= ["--tags=#{ENV['TAGS']}"]
```
Br3nda, line 94:
```suggestion
      raw_args <<= ['--tags="#{ENV['TAGS']}"']
```

Vagrantfile
```diff
@@ -89,6 +89,13 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         "redis.test"
       ]
     }
+    raw_args = []
+      raw_args << "--tags=#{ENV['TAGS']}" if ENV["TAGS"]
```
Br3nda, line 93: still feels like this needs escaping. Could it have had a namespace? e.g. we can call the variable `OAF_TAGS` or something, so we don't bring in variables that something else sets.

Br3nda: it's in full ruby code here, right? so a hash of key=>value, and we shell escape the value seems safer

ianheggie-oaf: Fixed in 1ac768de9e9a4bff7cac39e40a6d1945246a6ca3

ianheggie-oaf:
@Br3nda - updated based on your comments, please review and merge if approved

Br3nda (approved):

ianheggie-oaf:
@Br3nda - I have rebased on latest main - are you happy with this?

PR #220, opened 2025-09-23
Update Metabase from v0.47.9 to 0.56.6

benrfairless:
This pull request updates the Metabase service to use a newer version of the Metabase image in the `docker-compose.yml` file.

* Upgraded the Metabase Docker image from `v0.47.9` to `0.56.6` in `roles/internal/metabase/templates/docker-compose.yml` to ensure the service uses the latest features and security updates.

This can be deployed with Ansible without any customisation required on the server - refer to the [documentation](https://www.metabase.com/docs/v0.56/installation-and-operation/upgrading-metabase#what-happens-during-an-upgrade-or-downgrade) for confirmation.

### DIFF
--- a/roles/internal/metabase/templates/docker-compose.yml
+++ b/roles/internal/metabase/templates/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   metabase:
-    image: metabase/metabase:v0.47.9
+    image: metabase/metabase:0.56.6
     # Using 8000 as the port on the host to be consistent with planningalerts
     ports:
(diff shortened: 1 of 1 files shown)


Br3nda:
Have you deployed this before?

benrfairless:
> Have you deployed this before?
> 
> 

I have not, but I read the documentation which says this is all that should be required. 

This service is non-critical, internal only, and doesn't have write access to any production data.

Br3nda (approved):

PR #228, opened 2025-10-01
Reconfigure insecure google maps api key permissions

jattree:
Updates insecure client/js google api maps key permissions and add server side google key.


jamezpolley (approved):
LGTM.

PR #231, opened 2025-10-03
Add notes and a helper for running keybase headless

jamezpolley:
On a few occasions, I've run into problems running keybase on a headless system. Today it happens to be related to running inside WSL; last time it was running in a VM running on a remote machine that didn't have a desktop environment installed.

The problems I see have been annoyingly hard to diagnose. Things try to read from `/keybase` and just hang. It's often not obvious what's going on - especially because quite a lot of the time the thing that hands is shell autocompletion.

As a secondary issue - we know that the default symlinks (pointing to /keybase) don't work on OS X.

This changes does a few related things:

- the Makefile is updated to create `.keybase` if it doesn't already exist. It figures out where to point `.keybase` by asking Keybase where the magic keybase root is located - `keybase config get -d -b mountdir`. I haven't been able to test this on OS X yet (maybe @simonzippy or @benrfairless could check this for me?) but I think it should work there.
- symlinks are updated so that they now point at `.keybase` rather than `/keybase`. 
  - When the first git checkout happens the symlinks will be dangling as `.keybase` won't exist; but the first run of `make` will correct this.
- Adds `bin/headless-keybase.sh` as a helper script which sets up keybase to run headless.
  - Most of this is sourced from https://book.keybase.io/docs/linux#systemd-and-running-headless-keybase or elsewhere on that page
- Updates the README to describe these changes and how they're used.

I haven't documented this in the README, but part of my thinking here is that for cases where keybase can't be run, people in the past have done various things to change the symlinks to point to some other place, or even just dumped the secrets into the files directly and tried to remember to not check them in to git. They can now `mkdir .keybase`, do whatever they like to put the secrets in the right files in there, and voila - they now have a working config, with no problems caused by git seeing the symlinks having changed, much lower risk of checking the secrets in, etc.

PR #234, opened 2025-10-03
Prototype: use Hermit to simplifiy initial setup

jamezpolley:
_I'm going to flag this as a draft because it's has a few things that should be pulled out and landed on their own - eg #231. This is just to get some feedback about whether this seems like it's worth pursuing_

The goal here is to reduce the time it takes to check out the Infra repo and being able to push changes to production. 
- I've not attempted to address building local dev environments with Vagrant - that can be addressed another day
- I've not (yet) attempted to address setting up local dev environments for the projects. Hermit's support for Ruby doesn't seem to be great. So I'm not sure if Hermit can ever do more than I'm doing here.

This replaces the steps of:
- Make sure you've got python3.11 installed
  - No, not "at least 3.11", we mean "exactly 3.11", later versions don't work.
- Create a virtualenv
- Install the right version of Ansible
- Figure out how to download and install Terraform

The Makefile already did a good job of installing the right version of Ansible, but relied on the person reading the Makefile to get the right version of python installed to start with. Hermit now removes the need to pay attention to the python version.

I had initially tested out a smaller change, which still created a virtualenv and installed python packages into it. I thought this might be better for two reasons:
1.) Smaller change - if you're used to `.venv/bon/activate` or `.venv/bin/ansible-playbook` those would still work
2) I know that most IDEs detect a virtualenv and can work with it; the same might not be true of the Hermit environment.

But for what we're doing 2) isn't a big issue - all we're really installing is Ansible and its dependencies.

I've ended up going with this approach, which is still fairly familiar to anyone used to working with a virtualenv. You just need to `. bin/hermit-activate` rather than `. bin/activate`. After that, your path is set so that `pip` will run the right version and packages will be installed into `.hermit`. If you want to run the ansible binary without activating the environment, you can use `.hermit/python/bin/ansible` instead of `.venv/bin/ansible`

I don't plan to spend any more time on this unless it gets really good feedback. It was a fun way to spend an afternoon, and it did scratch an itch, but I think I've taken this about as far as it can go in this setup, and I'm not sure how much difference it makes.

### DIFF
--- a/roles/internal/righttoknow/templates/nginx/stage
+++ b/roles/internal/righttoknow/templates/nginx/stage
@@ -93,40 +93,40 @@ server {
     deny 8.222.0.0/16;
     deny 47.82.0.0/19;
 
-		# Additional ranges 2025-09-24
+    # Additional ranges 2025-09-24
+	
+	# Kingsoft Cloud
+    deny 104.250.52.0/22;
 		
-		# Kingsoft Cloud
-		deny 104.250.52.0/22;
+    # Meta / Facebook
+    deny 57.141.2.0/24;
 		
-		# Meta / Facebook
-		deny 57.141.2.0/24;
+    # Google Cloud & Crawlers
+    deny 34.174.0.0/17;
+    deny 66.249.68.0/24;
 		
-		# Google Cloud & Crawlers
-		deny 34.174.0.0/17;
-		deny 66.249.68.0/24;
+    # SEMrush SEO crawler
+    deny 185.191.171.0/24;
+    deny 85.208.96.0/24;
 		
-		# SEMrush SEO crawler
-		deny 185.191.171.0/24;
-		deny 85.208.96.0/24;
+    # Huawei Cloud
+    deny 202.76.160.0/20;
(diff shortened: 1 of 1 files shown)


jamezpolley:
> I was looking at possibly using [Mise](https://mise.jdx.dev) to set defaults and tasks across all our projects as standard.

I hadn't heard of Mise when I went down this rabbithole - but by chance, I came across its [use in one of our projects already ](https://github.com/openaustralia/morph/blob/a304cd13433a9fa6d9ab989151e39480cf135d62/provisioning/README.md?plain=1#L15), which makes me think it's worth looking at.

ianheggie-oaf:
> At first glance, Mise seems to have one big win over Hermit - it looks like it will not just replace the use of a python venv, but also replace use of rbenv, in addition to handling things like terraform.

I have moved to using mise on my local dev system and I like it.

ianheggie-oaf:
I think mise is best for me at least, but lets use `.ruby-version` and `.python-version`  as appropriate so we are agnostic to which tool on dev - I often then get Gemfile to read .ruby-version for consistency.

PR #243, opened 2025-10-14
Update Makefile with enhanced commands

benrfairless:
This pull request updates the `Makefile` and `README.md` to improve infrastructure management and clarify Keybase setup instructions, especially for Mac users. The changes introduce new Makefile targets for Terraform workflows, split Ansible commands for production and staging environments, and provide clearer documentation for Keybase symlink creation on different platforms.

**Makefile improvements:**

* Added new targets for Terraform (`tf-init`, `tf-plan`, `tf-apply`) to streamline infrastructure provisioning workflows.
* Split Ansible playbook commands into separate targets for production and staging environments for both checking (`check-rtk-prod`, `check-rtk-staging`) and applying changes (`apply-rtk-prod`, `apply-rtk-staging`), improving clarity and safety.
* Added a `macos-keybase` target to explicitly create the `.keybase` symlink for MacOS users.

**Documentation updates:**

* Updated `README.md` to clarify Keybase symlink setup, including instructions for Mac users to run `make macos-keybase` if needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: venv roles production ALL letsencrypt check-rtk
+.PHONY: venv roles production ALL letsencrypt
 
 ALL: venv roles .vagrant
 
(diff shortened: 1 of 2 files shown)


Br3nda (approved):
tested on ubuntu, works

Makefile
```diff
@@ -44,13 +44,31 @@ clean:
 clean-all: clean
 	rm -rf .vagrant
 
-check-righttoknow:
-	.venv/bin/ansible-playbook -i ./inventory/ec2-hosts site.yml -l righttoknow --check
+# Configure Keybase for MacOS
+macos-keybase:
+	ln -sf /Volumes/Keybase .keybase
```
jamezpolley: I don't *think* this should be needed.

I don't have a mac handy to test with, but my expectation was that the `keybase config get -d -b mountdir` command (line 6, this file) should work on a mac, and should return `/volumes/Keybase`. If that's right, this target shouldn't be needed - the same .keybase target should work on mac the same as it does on other *nix platforms.

I'm not objecting to this being added though - if it's needed, it's needed; if it's not strictly needed, it still adds a clearer alternative. I'm just surprised that the existing command didn't work.

jamezpolley (commented):
I would reject this due to the missing phony target,but it's already merged. I'll submit another PR to fix the phony target and add .keybase to the clean target.

... and now that I notice it - the clean target as well. And the new tf- targets.

PR #244, opened 2025-10-15
Tidy Makefile - dependencies, phony targets, clean .keybase

jamezpolley:
Several tidyups to the Makefile. I've grouped these together because they're all fairly straightforward

- Added all phony targets to .PHONY
- Tidied dependencies
    - lots of places referenced both venv and roles. Roles already depends on venv, so this isn't needed.
    - Lots of places now need to reference both .keybase and roles.
        - I considered making roles depend on .keybase. But there can be times when make roles is run and .keybase is not going to be needed - eg, if the user is just making roles to push into vagrant. So instead I'm setting $PRODUCTION and using that throughout the file.
- Remove the `ssh` target as it's been superseded. Also remove the deploy_user.yml file that's no longer needed.

Fixes identified in #243

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ retry: venv roles site.retry
 	.venv/bin/ansible-playbook site.yml -l @site.retry
 
 clean:
-	rm -rf .venv roles/external site.retry collections 
+	rm -rf .venv roles/external site.retry collections .keybase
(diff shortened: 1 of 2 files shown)


benrfairless (approved):
Looks OK to me

PR #245, opened 2025-10-15
Ignore terraform lock

jamezpolley:

PR #246, opened 2025-10-15
Use autovariables to reduce redundancy in rule definitions

jamezpolley:
depends-on: #245 

The intention here is to reduce redundancy in rule definitions, so that it's easier to ensure that any change that applies to all the related targets does apply to all of them.

The way it achieves this is via the $@ [auto-variable](https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html) so that we can have one rule with multiple targets, and run the command only on the target that was invoked.
 - Because the targets have prefixes (check-, apply-, tf-), [patsubst](https://www.gnu.org/software/make/manual/html_node/Text-Functions.html#index-patsubst-1) is used to trim the prefix. This is most clearly seen in the tf- rules - `$(patsubst tf-%,%,$@)`
- Additionally, I didn't want to break the use of "rtk" as shorthand for "righttoknow", so I've added a [subst](https://www.gnu.org/software/make/manual/html_node/Text-Functions.html#index-subst-1) call to handle that expansion.

 However, I'm not sure if the outcome is good. On the plus side - it does remove redundancy. On the minus side - it loses readability, and I'm not sure that the reduction in redundancy justifies the loss of readability.

Feedback requested!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,28 +45,16 @@ macos-keybase:
 	ln -sf /Volumes/Keybase .keybase
 
 # Terraform
-tf-init:
-	terraform -chdir=terraform init
-tf-plan:
-	terraform -chdir=terraform plan
-tf-apply:
-	terraform -chdir=terraform apply
+tf-init tf-plan tf-apply:
+	terraform -chdir=terraform $(patsubst tf-%,%,$@)
+
 
 # Checks only
-check-rtk-prod: $(PRODUCTION)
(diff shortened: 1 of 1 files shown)


ianheggie-oaf:
@jamezpolley - looks interesting, are you going to resolve conflictys then submit PR, or will it stay in Draft limbo?

benrfairless:
This has been in draft for some time, so I'm going to close it for now. If we need it we can always re-open it.

PR #256, opened 2025-10-28
git ignore the secrets file

Br3nda:
## What does this do?
tells git to ignore the secrets file


benrfairless (changes requested):
I would have thought that this line is enough: https://github.com/openaustralia/infrastructure/blob/d85cd41eb560024cf9a1ea20e91a2c4ed541c240/.gitignore#L10

If not, can you please move your change further up.

PR #252, opened 2025-10-23
Update Ansible Configuration for new Right to Know Production Server

benrfairless:
This pull request introduces a new standalone production environment for Right to Know, reorganizes inventory and group variable files to support this, and updates deployment, certificate, and cron job logic to distinguish between production and staging environments more clearly. The changes improve separation and clarity between staging and production, enabling more granular management and deployment.

Key changes include:

**Infrastructure and Inventory Updates:**

* Added `prod.righttoknow.org.au` to the `inventory/ec2-hosts` file and updated inventory group names to use underscores (e.g., `[righttoknow_production]`, `[righttoknow_staging]`), aligning with Ansible best practices and supporting the new standalone production host. [[1]](diffhunk://#diff-c54ef71957ec91854593c9dc0e94146184afc872739b098c66473860e556f2cbR6) [[2]](diffhunk://#diff-c54ef71957ec91854593c9dc0e94146184afc872739b098c66473860e556f2cbR27-R32)

* Created a new `group_vars/righttoknow_production.yml` file with production-specific secrets, domain settings, and log configurations for the new standalone production environment.

**Makefile and Deployment Process:**

* Refactored Makefile targets to use new names (`check-righttoknow-prod`, `apply-righttoknow-staging`, etc.) that match the updated inventory groups and hostnames, ensuring deployment commands target the correct environments.

**Certificate Management:**

* Updated the certificate issuance logic in `roles/internal/righttoknow/tasks/certificates.yml` to differentiate between standalone staging and production, assigning the correct domains and certificate sets based on group membership and environment flags.

**Cron Job Configuration:**

[...]

PR #257, opened 2025-10-28
Multiple Ansible Updates to provision new servers

benrfairless:
## What does this do?
Makes the below changes:
- Disable python installation task for the moment as it doesn't work with new server builds
- Rename pip installation tasks for clarity and specify conditions based on Ubuntu releases
- Update ansible_user to 'ubuntu' in update_ubuntu_keys.yml task
- Update ansible_ssh_private_key_file from 'test.pem' to 'terraform.pem' for ec2 instances.
- Ensure all tasks in deploy-user ansible role are run with elevated permissions.

## Why was this needed?
These changes have been split from #252 as they affect Ansible provisioning across the board.

No deployment steps required

These changes can be auto-merged once we are happy with them

### DIFF
--- a/group_vars/ec2.yml
+++ b/group_vars/ec2.yml
@@ -6,7 +6,7 @@ oaf_domain: oaf.org.au
 opengovernment_domain: opengovernment.org.au
 electionleaflets_domain: electionleaflets.org.au
 ansible_user: ubuntu
-ansible_ssh_private_key_file: 'test.pem'
+ansible_ssh_private_key_file: 'terraform.pem'
 ec2_region: 'ap-southeast-2'
 # TODO: We only need this key to have limited privileges. It currently has too many privileges.
 aws_access_key: !vault |
--- a/roles/internal/deploy-user/tasks/main.yml
+++ b/roles/internal/deploy-user/tasks/main.yml
@@ -1,8 +1,10 @@
 - name: Ensure we have a deploy group
+  become: true
   group:
     name: deploy
 
 - name: Ensure we have a deploy user
+  become: true
   user:
     name: deploy
     comment: "Deploy user"
@@ -13,6 +15,7 @@
 ## but it does not remove other keys. If we have a need to revoke
 ## access we'll have to do something different.
 - name: Create authorized_keys files for local accounts
+  become: true
   authorized_key:
     user: "{{ item[0] }}"
     key: https://github.com/{{ item[1] }}.keys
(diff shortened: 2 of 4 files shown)


roles/internal/deploy-user/tasks/main.yml
```diff
@@ -1,8 +1,10 @@
 - name: Ensure we have a deploy group
+  become: true
```
jamezpolley: I might be wrong, but I thought that all the tasks in this role should have become set already. Not sure where that happens though, and having them listed here explicitly certainly makes it more obvious.

jamezpolley (dismissed):

update_ubuntu_keys.yml
jamezpolley: I've raised #258 which removes this file entirely.

benrfairless (commented):
> I might be wrong, but I thought that all the tasks in this role should have become set already. Not sure where that happens though, and having them listed here explicitly certainly makes it more obvious. 

When I run this without the ```become:``` line set, I get an error that I'm not running it as sudo (which is accurate, as ansible logs into the server as ubuntu, not root).

group_vars/ec2.yml
```diff
@@ -6,7 +6,7 @@ oaf_domain: oaf.org.au
 opengovernment_domain: opengovernment.org.au
 electionleaflets_domain: electionleaflets.org.au
 ansible_user: ubuntu
-ansible_ssh_private_key_file: 'test.pem'
+ansible_ssh_private_key_file: 'terraform.pem'
```
Br3nda: why did the file name changes?

update_ubuntu_keys.yml
```diff
@@ -18,6 +18,6 @@
 
 - hosts: ec2
   vars:
-    ansible_user: root
+    ansible_user: ubuntu
```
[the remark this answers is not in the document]
benrfairless: This change is still needed if we decide to keep this file.

benrfairless:
> do we have a good place to add a fix for this? New servers need this folder to be writable by the `deploy` user
> 
> 
> 
> ```
> 
>  ** [prod.righttoknow.org.au :: err] fatal: could not create leading directories of '/srv/www/production/shared/cached-copy': Permission denied
> 
> ```

This should already happen [here](https://github.com/openaustralia/infrastructure/blob/main/roles/internal/righttoknow/tasks/main.yml#L81-L89)

benrfairless:
I've reviewed and updated comments for the various changes. Hopefully this addresses your concerns?

Br3nda (approved):
happy for this to be applied. and if further changes are needed, we'll make more PRs.

we're marching onwards out of ubuntu 18, and like @jamezpolley said we're super unlikely to want to build a new ubuntu18.

PR #258, opened 2025-10-28
Remove update_ubuntu_keys.yml playbook as it's no longer needed.

jamezpolley:
This mirrors dd3eaa3 in which a similar file was already removed. This file duplicates the functionality in that file, and is being removed for the same reason - it's been made redundant by the same better solution. dd3eaa3 was merged in #244 

## Notes to reviewer

I've had a quick check to see if there's anything that calls this, or any docs that refer to it. I can't find any - if there are some, they should be cleaned up as part of this PR as well.


benrfairless (approved):
Perhaps updating the doco to explain how to refresh the keys would be helpful.

Br3nda (approved):

PR #259, opened 2025-10-28
RTK: Increase Root volume size to 20GB

benrfairless:
## Relevant issue(s)
N/A

## What does this do?
Sets the default root volume size on righttoknow-production and righttoknow-staging to 20GB from their default of 12GB.

## Why was this needed?
12GB is not enough for the tasks these servers need to perform. 20GB ensures the disk doesn't run own of space.

Nil

Nil


jattree (approved):
Looks good!

PR #275, opened 2025-11-11
Morph Terraform Improvement

benrfairless:
## Relevant issue(s)
N/A

## What does this do?
- Enabled automatic backups for the Linode instance by setting `backups_enabled` to `true` in `main.tf`.
- Removed resize Disk Option from main.tf

## Why was this needed?
These changes have been manually made in production without updaitng terraform. 

No steps required

### DIFF
--- a/terraform/morph/main.tf
+++ b/terraform/morph/main.tf
@@ -16,7 +16,7 @@ resource "linode_instance" "main" {
   label            = "morph"
   group            = "morph"
   booted           = true
-  backups_enabled  = false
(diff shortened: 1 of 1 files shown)


benrfairless:
This is a change to the terraform infrastructure which I don't believe would be covered off in https://github.com/openaustralia/morph/pull/1398. We shouldn't have multiple instances of terraform running in different repos for the same resource.

jattree (approved):
Looks good

Issue #241, opened 2025-10-14
stouts.backup role does not work on Ubuntu 22.04


benrfairless:
**Describe the bug**
The stouts.backup role does not work on Ubuntu 22.04.

**To Reproduce**
1. Enable stouts.backup on a server that is running Ubuntu 22.04
2. Attempt to deploy via Ansible

**Expected behavior**
Stouts.backup should be installed

**Additional context**
This is caused by [this dependency](https://github.com/Stouts/Stouts.backup/blob/37c76278482e29744ef894abed62d399e4aff245/tasks/install.deb.yml#L10) (python-boto) which doesn't exist on 22.04

PR #278, opened 2025-11-12
Updates the Stouts.backup role condition to support both Ubuntu 18.04

jattree:
(bionic) and Ubuntu 22.04 (jammy). This enables automated backups to S3 for Right to Know servers running on Ubuntu 22.04.

## Relevant issue(s)
* https://github.com/openaustralia/infrastructure/issues/241

### DIFF
--- a/roles/internal/righttoknow/meta/main.yml
+++ b/roles/internal/righttoknow/meta/main.yml
@@ -140,7 +140,7 @@ dependencies:
         target: "s3://s3.amazonaws.com/oaf-backups/righttoknow/{{ inventory_hostname }}/data"
         exclude:
           - "*/bundle"
-    when: ansible_distribution_release == 'bionic'
+    when: ansible_distribution_release in ['bionic', 'jammy']
     
   - role: nickhammond.logrotate
     logrotate_scripts:


benrfairless:
I've tried running `ansible-playbook site.yml -l righttoknow_production -C -D` and I get this error

```ansible
TASK [Stouts.backup : Install dependencies] ***********************************************************************************************************************************************
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying `name: "{{item}}"`,
please use `name: ['cron', 'gzip', 'python-boto', 's3cmd']` and remove the loop. This feature will be removed from ansible-base in version 2.11. Deprecation warnings can be disabled by
setting deprecation_warnings=False in ansible.cfg.
failed: [prod.righttoknow.org.au] (item=['cron', 'gzip', 'python-boto', 's3cmd']) => {"ansible_loop_var": "item", "changed": false, "item": ["cron", "gzip", "python-boto", "s3cmd"], "msg": "No package matching 'python-boto' is available"}
```

After discussion with @jattree it seems that this is because the check function doesn't work as expected. Jon has confirmed that backups are actually running, so I'm going to approve/merge this.

benrfairless (approved):

PR #280, opened 2025-11-13
Add Ubuntu 22.04 support to Stouts.backup role

### LINKED ISSUE #241: stouts.backup role does not work on Ubuntu 22.04

jattree:
Updates the Stouts.backup role's install.deb.yml to support Ubuntu 22.04 (jammy) by adding a separate task that installs python3-boto3 instead of python-boto for focal and jammy distributions.

This fixes the package installation error where python-boto is not available on Ubuntu 22.04 systems.


Br3nda (approved):

benrfairless (changes requested):
Hold on this please. I'll update shortly

PR #287, opened 2025-11-20
Fix/chrome check mode

Br3nda:
when in check mode, ansible errors on downloading latest chrome. Changed to not run this in check mode


benrfairless (approved):
I would have preferred if we used `check_mode: no` but I tried that and it didn't work.

Hence this is now approved 👍

PR #293, opened 2025-11-24
Fork Stouts.backup to internal oaf.backup role with Python 3 support

jattree:
- Fork Stouts.backup (v3.4.2) to roles/internal/oaf.backup
- Add python3-boto3 support for Ubuntu 20.04 (focal) and 22.04 (jammy)
- Update 5 roles to use oaf.backup: righttoknow, openaustralia, oaf, electionleaflets, opengovernment
- Remove Stouts.backup from requirements.yml
- Remove incorrectly tracked file from roles/external/

The upstream Stouts.backup role (last updated 2016) installs python-boto which is not available on Ubuntu 20.04+. This fork adds conditional installation of python3-boto3 for newer Ubuntu versions.

Forking to roles/internal/ rather than modifying roles/external/ because external roles are managed by ansible-galaxy and any local changes would be overwritten when running `ansible-galaxy install -r requirements.yml`.

## What does this do?
Forks stout backup function that forces python 2.x

## Why was this needed?
upstream decencies have no support for python3 and patching the external files not clean.


roles/internal/oaf.backup/.travis.yml
```diff
@@ -0,0 +1,33 @@
+---
```
Br3nda: We don't have travis.. did this come from upstream?

roles/internal/oaf.backup/tasks/backup.yml
```diff
@@ -0,0 +1,13 @@
+---
+
+- include: install.deb.yml
+  when: ansible_os_family == "Debian" and not backup_remove
+
+- include: install.red.yml
+  when: ansible_os_family == "RedHat" and not backup_remove
```
Br3nda: do we have any redhat machines?

Br3nda (approved):

PR #300, opened 2025-12-01
Update DNS records to point test. at wp.com hosted blog

jamezpolley:
Change DNS records to use the new wp.com hosted site for test.

Once we start using th is site, we'll change things so the test. name goes to a new staging site we set up there, and the main www records will redirect to the wp.com hosting


Br3nda (approved):

Br3nda:
let's not run until we're happy with the state of our new blog.. but .... sooooooon

benrfairless (approved):

jamezpolley:
> let's not run until we're happy with the state of our new blog.. but .... sooooooon

Checked this with @Br3nda offline, confirmed that we're both okay with updating the test records now but not the main records. moving ahead on that basis.

PR #304, opened 2025-12-02
Check terraform modules are updated/checked in

Br3nda:
## What does this do?

in CI, run `make tf-init` and then check if this changed the `terraform/.terraform.lock.hcl` from what's in git.

Fail if it changed.


Br3nda:
needs some secrets present to fully run this, so I'm abandoning this idea

PR #311, opened 2025-12-05
Block bot putting load on services

Br3nda:
## Relevant issue(s)
Bot hitting TVFY today, creating high load and using all resources

## What does this do?
Block that range

Same range is already blocked on RTK

### DIFF
--- a/roles/internal/theyvoteforyou/templates/production
+++ b/roles/internal/theyvoteforyou/templates/production
@@ -24,7 +24,9 @@ server {
     deny 47.82.0.0/19;
     deny 47.82.12.0/23;
     deny 47.82.14.0/23;
-
+    # Alibaba Cloud
+    deny 47.79.192.0/19;
+    
     # Desperate measures - blocking whole ASN 136907
     # This list from https://www.enjen.net/asn-blocklist/index.php?asn=136907&type=nginx
     deny 101.46.136.0/21;


benrfairless (approved):

roles/internal/theyvoteforyou/templates/production
```diff
@@ -24,7 +24,9 @@ server {
     deny 47.82.0.0/19;
     deny 47.82.12.0/23;
     deny 47.82.14.0/23;
-
+    # Alibaba Cloud
+    deny 47.79.192.0/19;
```
jamezpolley: > Same range is already blocked on RTK

whois suggests this is part of a much larger block

```
NetRange:       47.74.0.0 - 47.87.255.255
CIDR:           47.74.0.0/15, 47.76.0.0/14, 47.80.0.0/13
NetName:        AL-3
NetHandle:      NET-47-74-0-0-1
Parent:         NET47 (NET-47-0-0-0-0)
NetType:        Direct Allocation
OriginAS:
Organization:   Alibaba Cloud LLC (AL-3)
RegDate:        2016-03-17
Updated:        2017-04-26
Ref:            https://rdap.arin.net/registry/ip/47.74.0.0
```

So if I was going to do this fresh I might suggest a larger block. But given that we're copying from RTK, I like the idea of just going with what has worked there.

jamezpolley (commented):
LGTM

PR #315, opened 2025-12-05
Fix OpenVPN client config and IAM auth script

jattree:
What does this do?

  - Fixes OpenVPN client config template to embed CA certificate and TLS auth key inline
  - Updates IAM auth script to use /usr/local/bin/aws (pip-installed) instead of system aws
  - Installs awscli via pip instead of apt to avoid Python package version conflicts
  - Adds existing vpn-users IAM group membership to Terraform state (jon, ben, brenda, ian+oaf)

  Why was this needed?

  - Client config was missing embedded certificates, causing "You must define CA file" error on connection
  - System awscli (apt) had botocore version conflicts causing auth script to fail with Python tracebacks
  - VPN users group membership was manually configured but not tracked in Terraform

  Implementation/Deploy Steps (Optional)

  1. Terraform: terraform apply to create the group membership resource (users already in group, just adding to state)
  2. Ansible: Re-run playbook to regenerate client config with embedded certs: ansible-playbook site.yml -l openvpn
  3. Distribute new client-template.ovpn to users

  Note: Server already has these fixes applied manually during troubleshooting.

  Notes to reviewer (Optional)

  VPN is currently working with these changes applied directly to the server. This PR captures those fixes in code.

### DIFF
--- a/roles/internal/openvpn/tasks/main.yml
+++ b/roles/internal/openvpn/tasks/main.yml
@@ -8,14 +8,15 @@
       - easy-rsa
       - python3-pip
       - jq
-      - awscli
     state: present
     update_cache: yes
     cache_valid_time: 3600
 
-- name: Install boto3 for IAM authentication
+- name: Install AWS CLI and boto3 via pip
   pip:
-    name: boto3
+    name:
+      - boto3
+      - awscli
     executable: pip3
     state: present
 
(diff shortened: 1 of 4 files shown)


benrfairless (approved):
Approving this as it’s already live.

When we add @jamezpolley (tchaypo) to the IAM group there will need to be another update to the vpn-iam.tf file.

PR #317, opened 2025-12-05
Add per-service security groups with Cloudflare IP restrictions

jattree:
Relevant issue(s)

  N/A

  What does this do?

  - Creates dedicated security groups for theyvoteforyou, righttoknow, oaf, openaustralia, opengovernment (identical rules to
  shared webserver SG)
  - Attaches both webserver and new per-service SGs to instances (transition phase)
  - Adds dynamic Cloudflare IP fetching from cloudflare.com/ips-v4 and cloudflare.com/ips-v6
  - Adds *_cloudflare_only variables (default false) for each service to enable Cloudflare IP restrictions
  - Switches modules to use vpc_security_group_ids for in-place SG updates without instance replacement

  Why was this needed?

  The shared webserver security group prevents per-service Cloudflare proxy enablement. With dedicated SGs, we can:
  1. Enable Cloudflare proxy per-service by setting <service>_cloudflare_only = true
  2. Restrict HTTP/HTTPS traffic to Cloudflare IPs only on a per-host basis
  3. Gradually migrate services behind Cloudflare without affecting others

  Implementation/Deploy Steps

  1. terraform plan - verify 5 new SGs created, 6 instances updated in-place (no replacements)
  2. terraform apply - creates SGs and attaches them alongside existing webserver SG
  3. Test all services still accessible
  4. Future PR: Enable Cloudflare per-service and remove webserver SG

  Notes to reviewer

  - Cloudflare IPs are fetched dynamically at plan/apply time
  - All instance updates are in-place (no downtime)
  - The cloudflare_only variables are all false by default - no traffic restrictions applied yet
  - electionleaflets S3 bucket destroy in plan is a pre-existing issue, unrelated to this PR


Br3nda (approved):

benrfairless (dismissed):
Can you please double check that Port 25 is enabled on Right to Know? I see that `sg-5f58e926 - incoming_email` will be removed but I can't see it being added in any other group.

benrfairless (approved):

PR #321, opened 2025-12-09
TheyVoteForYou staging using mysql 8

Br3nda:
## What does this do?

Reconfigures TVFY staging to use a new database that runs on mysql8

### DIFF
--- a/roles/internal/theyvoteforyou/tasks/main.yml
+++ b/roles/internal/theyvoteforyou/tasks/main.yml
@@ -42,7 +42,7 @@
 
 - name: Create theyvoteforyou databases
   mysql_db:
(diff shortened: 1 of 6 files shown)


roles/internal/theyvoteforyou/tasks/main.yml
```diff
@@ -62,7 +62,7 @@
 
 - name: Create mysql user with readonly access to the production theyvoteforyou database for metabase
   mysql_user:
-    login_host: "{{ mysql_5_host }}"
+    login_host: "{{ mysql_8_host }}"
```
Br3nda:
```suggestion
    login_host: "{{ mysql_5_host }}"
```

benrfairless (approved):
LGTM

PR #330, opened 2025-12-30
stop replicating the OAF blog

Br3nda:
We're not migrating this to MySQL 8. Instead we're moving to wordpress.com

### DIFF
--- a/terraform/dms-migration.tf
+++ b/terraform/dms-migration.tf
@@ -57,18 +57,18 @@ locals {
         rule-action = "include"
       }
     ] : [],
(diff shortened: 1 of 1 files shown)


terraform/dms-migration.tf
```diff
@@ -57,18 +57,18 @@ locals {
         rule-action = "include"
       }
     ] : [],
```
Br3nda:
```suggestion
    ] : []
```

benrfairless (approved):
LGTM

PR #331, opened 2025-12-30
Move TVFY to mysql 8

Br3nda:
## What does this do?

1. Stops replicating from mysql5 host to mysql8 host
1. Point TVFY at the MySQL8 host

Note: Staging is already using this host


benrfairless (approved):
LGTM. Merge and deploy when you're ready.

benrfairless (approved):

PR #336, opened 2026-01-03
Update logrotate configuration to use systemctl for rsyslog restart

benrfairless:
## What does this do?
Updates the logrotate configuration to restart rsyslog using systemctl instead of the previous method.

## Why was this needed?
At the moment, the current setup doesn't restart rsyslog correctly which results in files not being written to the correct day.


jattree (approved):
Looks perfect!

PR #338, opened 2026-01-06
Update Nginx Deny list to block SQL Injection attempts

benrfairless:
## Relevant issue(s)
- N/A

## What does this do?
Updates the Nginx configuration to deny access from an IP address associated with SQL injection attempts.

## Why was this needed?
We received numerous alerts that this IP address was attempting to perform SQL injection using the search function in Right to Know.

This has already been deployed as an emergency change

This was affecting production hence the change was deployed urgently.


ianheggie-oaf (approved):
Makes sense to me

PR #339, opened 2026-01-11
adding a new server for running soon-to-be-upgraded openaustralia.org.au

Br3nda:
## Relevant issue(s)
to get openaustralia.org.au to connect to mysql8, it needs upgrading to php7 or php8
The PHP 7 on the current 16.04 server doesn't support modern password auth that MySQL 8 in AWS RDS demands. 
## What does this do?
This shaves one yak. Specifically it gives us a new server to put staging on, and php8, and connect to MySQL8

The PHP upgrade is underway here:
https://github.com/openaustralia/twfy/pull/44

## Why was this needed?
The code is 19 years old. It's time.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -100,6 +100,7 @@ module "openaustralia" {
   security_group_service   = aws_security_group.openaustralia
   instance_profile         = aws_iam_instance_profile.logging
   ami                      = var.ubuntu_16_ami
+  ubuntu_24_ami            = var.ubuntu_24_ami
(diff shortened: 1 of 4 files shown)


benrfairless (approved):
@Br3nda I have updated this PR to use the same standards as the new RTK server. 

I'm assuming Ubuntu 24 with extended root drive to 20GB.

If you need I can create a DNS entry but I haven't done that as yet.

Otherwise this is approved.

Issue #312, opened 2025-12-05
[EPIC] Migrate our services behind Cloudflare WAF


benrfairless:
**Is your feature request related to a problem? Please describe.**
Several services continue to experience problems with high load that is generated from automated requests. At present the only way we are resolving these issues is by manually applying deny rules to large IP ranges.

**Describe the solution you'd like**
We need to implement a proper WAF in front of our services. Thankfully Cloudflare provides a free business plan to each of our domains.

**Describe alternatives you've considered**
Other solutions either require local maintenance or incur costs. Given that we have access to a business grade offering for free, it makes sense that we should use it.

**Additional context**
This is a master issue to capture the reason and context behind the decision to implement Cloudflare and can be used for discussion around the concept in general. A separate issue will be created for each service that needs to be migrated.

Issue #169, opened 2024-10-09
Consider changing to CloudFlare SSL


benrfairless:
----------
Suggestion from user below:
----------

I see that you use Let's Encrypt for the websites SSL certificate, which requires renewing every 90 days. As you are aware, this causes brief outages.

I would recommend using CloudFlare's SSL certificates, as they auto renew. It is completely free. There is a reason almost 20% of the entire internet use CloudFlare's services (most of them on the free plan).

--
Source: https://mail.missiveapp.com/#team_unassigned/8555ecd2-4304-41e5-ab2d-17e392eab5df_team_unassigned/conversations/301b878a-33f3-4d29-b93b-bf15d5ec4e57

benrfairless:
In order for this to work, we would need to:

- [ ] Order a [Universal SSL](https://developers.cloudflare.com/ssl/edge-certificates/universal-ssl/enable-universal-ssl/) from Cloudflare
- [ ] Update DNS to proxy the DNS traffic through Cloudflare
- [ ] Follow the process [here](https://developers.cloudflare.com/ssl/origin-configuration/origin-ca/#_top) to create and install the Origin CA Certificate on the RTK server
- [ ] Change the SSL/TLS Mode to Full (strict).
- [ ] Remove certbot and LetsEncrypt

PR #355, opened 2026-02-03
Create new role to support Cloudflare Proxy

benrfairless:
## Relevant issue(s)
- https://github.com/openaustralia/infrastructure/issues/312
- https://github.com/openaustralia/infrastructure/issues/169

## What does this do?
Adds a role to configure nginx and apache to restore real client IP addresses when behind a Cloudflare proxy.

## Why was this needed?
When Cloudflare proxies requests, the original client IP is replaced with a Cloudflare IP. This role ensures that the real client IP can be restored using the appropriate headers.

- No Deployment required at this time.

- This is a new role which is not currently applied to any service.


docs/cloudflare-proxy-migration.md
```diff
@@ -0,0 +1,235 @@
+# Cloudflare Proxy Migration Guide
+
+This document covers the infrastructure changes required when migrating OAF services to sit behind Cloudflare's proxy (orange cloud mode).
+
+## Overview
+
+When Cloudflare proxy is enabled for a domain, all HTTP/HTTPS traffic flows through Cloudflare's edge network before reaching our origin servers. This provides DDoS protection, caching, and WAF capabilities, but requires configuration changes to:
+
+1. Restore real client IP addresses in nginx and apache
```
jamezpolley: First thought was that "restore" referred to "change config to make nginx/apache serve on real IPs". But that didn't make sense. Have re-read - it mentions client IP. I think that this is talking about grabbing the original client IP out of xomething like the X-FORWARDED-FOR header and making sure that that's what we log, right?

If that's the case - maybe something like "Adjust Nginx/Apache to log the original client IP from X-HEADER" would be more clear?

docs/cloudflare-proxy-migration.md
```diff
@@ -0,0 +1,235 @@
...
+
+Setting these to `true` creates AWS security group rules that only allow HTTP/HTTPS from Cloudflare IP ranges (dynamically fetched from `cloudflare.com/ips-v4` and `ips-v6`).
+
+**Files:**
+- [terraform/cloudflare-security-groups.tf](../terraform/cloudflare-security-groups.tf)
+- [terraform/variables.tf](../terraform/variables.tf)
+
+### Ansible (New Role)
+
+The `oaf.cloudflare_realip` role configures nginx or apache to restore real client IPs:
```
jamezpolley: Similar to my first comment about "restore real IP" (but more minor, if the first comment is addressed I'm not as fussed about whether this is changed)

If it's not difficult to change, I think something like "log_cloudflare_client_ip" would be more clear about what this does. It's long though. Maybe just "log_header_ip"?

docs/cloudflare-proxy-migration.md
jamezpolley: Ah - reading below - it's the `CF-Connecting-IP` header.

Br3nda (approved):

PR #372, opened 2026-03-03
Update Right to Know Production to use Cloudflare

benrfairless:
## Relevant issue(s)
- https://github.com/openaustralia/infrastructure/issues/312

## What does this do?
1. Updates Terraform for the root and www domains on Right to Know to enable the proxy (orange cloud).
2. Adds the cloudflare_realip role as a dependency of the righttoknow role so that the real IP address is passesd through to Right to Know.

## Why was this needed?
This change ensures that the Nginx webserver can correctly handle real IP addresses from Cloudflare and improves the DNS configuration by enabling proxying for better performance and security.

- Run `make tf-plan` to verify changes to be applied via Terraform
- Run `make check-righttoknow-all` to verify changes to be applied via Ansible
- When satisifed with both, run `make tf-apply` and `make apply-righttoknow-all`

- I will be implementing this change once it is approved.

### DIFF
--- a/roles/internal/cloudflare_realip/tasks/main.yml
+++ b/roles/internal/cloudflare_realip/tasks/main.yml
@@ -13,6 +13,7 @@
   delegate_to: localhost
   become: false
   run_once: true
+  check_mode: false
 
 - name: Fetch Cloudflare IPv6 ranges
   uri:
@@ -22,6 +23,7 @@
   delegate_to: localhost
   become: false
   run_once: true
+  check_mode: false
   when: cloudflare_include_ipv6
 
 - name: Parse Cloudflare IP ranges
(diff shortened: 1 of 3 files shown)


roles/internal/cloudflare_realip/tasks/main.yml
```diff
@@ -22,6 +23,7 @@
   delegate_to: localhost
   become: false
   run_once: true
+  check_mode: false
```
Br3nda: why don't we run this in check mode?

Br3nda (commented):
not sure why we aren't checking the ipv4/6 ip in check mode

Br3nda (approved):

PR #374, opened 2026-03-10
Update user authentication flow

benrfairless:
## Relevant issue(s)
- N/A

## What does this do?
Fix an issue preventing access to the Sidekiq console on Right to Know

## Why was this needed?
We neeed to pass requests directly to the back-end, as sesssion cookies must be preserved and responses should not be cached (even by Varnish).

- Deploy by running `make apply-righttoknow-all`

- Manual fix has already been deployed in production

### DIFF
--- a/roles/internal/righttoknow/files/default.vcl
+++ b/roles/internal/righttoknow/files/default.vcl
@@ -68,6 +68,12 @@ sub vcl_recv {
         return (pass);
     }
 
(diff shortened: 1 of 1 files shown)


ianheggie-oaf (approved):
Approved - Since its already in production, you have presumably already smoke tested it.

I note that https://www.righttoknow.org.au/sidekiq comes back as follows when I am signed in and when not - is this expected?
> Sorry, we couldn't find that page
> The page doesn't exist.

PR #378, opened 2026-03-16
Update Metabase Docker image to version 0.59.2

benrfairless:
## Relevant issue(s)
- N/A

## What does this do?
- Updates the Metabase Docker image reference to use a variable
- Updates Metabase to the latest version (0.59.2).

## Why was this needed?
- To ensure the deployment uses the latest version of Metabase
- Make it easier to make future upgrades by using a variable for the image reference

- run `make apply-metabase` from repo

- There is a (minor) chance that the disk might fill up when the ansible apply is run. If this is the case, a manual change would be to remove any old docker images to free up disk space.


ianheggie-oaf (approved):
Approved given the metabase role is only used for sites in group metabase in sites.yml

PR #387, opened 2026-03-30
Remove unnecessary shared files from configuration

benrfairless:
## Relevant issue(s)
- Nil

## What does this do?
Removes unnecessary shared files from the configuration in `general.yml`.

## Why was this needed?
Required to ensure that capistrano deployments continue to work into the future. Also streamlines the configuration by removing files that aren't required to be linked because they don't exist.

`make apply-righttoknow-all` to implement these changes

This is a minor change which prevents


ianheggie-oaf (approved):
Makes sense to me

PR #402, opened 2026-04-13
Tasks for removing mise or rvm (into main, already in new)

Br3nda:
## What does this do?
Makes tasks available to remove RVM or mise

## Why was this needed?
we have multiple tools on some machines.


roles/internal/remove_rvm/README.md
```diff
@@ -0,0 +1,31 @@
+# `remove_rvm`
+
+Undoes a system-wide RVM installation as performed by `roles/external/rvm.ruby`
```
benrfairless, line 3: Does the external role have the ability to uninstall RVM?

benrfairless (changes requested):
Please ensure that new roles start with `oaf.` and do not use an underscore.

ianheggie-oaf:
@benrfairless and @Br3nda - FYI the rvm.ruby role can remove a specific version of ruby from rvm, but not rvm itself.

Noting this PR is for merge of my work (that is already in new-openaustralia-ansible) into main, SO the recommendation above probably should be in a part2 PR (I was at the time focused on the min needed to sort out the mess of multiple version managers fighting - with env having a mix of both, not on changing how we trigger roles to run)

Br3nda:
@ianheggie-oaf I was trying to explain how the parser finds its ruby. The new bin run thing... But also the Ruby it wants needs to added here in the infrastructure repo. Does that sound correct?

PR #403, opened 2026-04-13
deploy user needs to be in the rvm group

Br3nda:
## What does this do?

Adds the deploy user to the "rvm" group so it can use rvm


benrfairless (changes requested):
If you run this on a server that doesn't have rvm, you get this error:
```
Group rvm does not exist
```

This will also be a problem if rvm is not installed as this runs on all servers before an app specific tasks.

ianheggie-oaf:
> If you run this on a server that doesn't have rvm, you get this error:
> 
> ```
> Group rvm does not exist
> ```
> 
> This will also be a problem if rvm is not installed as this runs on all servers before an app specific tasks.

@brenda - I agree - ensuring user exists should be separate from ensuring user is in group rvm - is there a group membership that says.

I suggest we set as appropriate in group_vars:
```
ruby_manager to 'rvm' or 'rbenv'
ruby_version to 3.3.4 or 3.4.4 (planningalerts and theyvoteforyou respectively)
rvm1_rvm_version (differs)
rvm1_gpg_key_server (differs)
```

And move the meta tasks preferably to site.yml

- run rvm.ruby role if ruby_manager = 'rvm' 
- run zzet.rbenv rile if ruby_manager = 'rbenv'

```
when: ruby_manager == 'rbenv'
```

I went to look at how to rvm or not to rvm is decided and found some of it is in meta requirements (Not a fan of this - I prefer roles to be explicit in base site.yml, but that is a red herring), anyway we have some with rvm, some with rbenv and personally I prefer mise:

```
$ ag rvm.ruby
roles/internal/planningalerts/meta/main.yml
3:  - role: rvm.ruby

roles/internal/theyvoteforyou/meta/main.yml
3:  - role: rvm.ruby

roles/internal/remove_rvm/README.md
3:Undoes a system-wide RVM installation as performed by `roles/external/rvm.ruby`
31:* The role assumes `rvm.ruby` was run as root, ie a system install

$ ag remove_rvm
roles/internal/openaustralia/meta/main.yml
125:  - role: remove_rvm
```

BTW - openaustralia-parser/bin/run  
checks and uses (in order):
* local rbenv (set up via /etc/profile.d/rbenv.sh on opoenserver (newprod()
* system rbenv
* mise
* local rvm
* system rvm (set up by /etc/profile.d/rvm.sh on PA)

and twfy/bin/run just uses system packages for php and perl

ianheggie-oaf:
@Br3nda OOPS - I accidentality rebased this on new-openaustralia-ansible as part of incorporating this on my fix (which was refactored significantly) - The following pull request makes this obsolete on new:
* https://github.com/openaustralia/infrastructure/pull/412

I have a suggested alternative on branch fix/deploy-user-addition-to-rvm-group if you need something for main.

PR #404, opened 2026-04-13
stop replicating mysql 5 to 8

Br3nda:
## What does this do?
turns off replication from mysql 5 to 8

Instead i'll make a manual copy of current prod to new prod.


benrfairless (approved):
Approved. I don't think this deletes the DB, just the DMS replication.

PR #408, opened 2026-04-13
Fix/merge main into new openaustralia ansible [Review 1st]

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/406

## What does this do?

Works through the conflicts between main and new-openaustralia-ansible branches

## Why was this needed?

because there where conflicts, and
So @Br3nda can see the changes

Issue #407, opened 2026-04-13
Incorporate manual changes to newprod back into ansible playbooks


ianheggie-oaf:
We have done a decent amount of manual configuration to get the newprod server working. This issue covers the work to fold it back in and then check its done by  provisioning a vagrant VM

ianheggie-oaf:
Closed - I merged what I saw with check, @Br3nda  and I pair reviewed and it has been deployed

PR #409, opened 2026-04-13
Simplify openaustralia ansible groups and ensure mysql DB 8.x as well as 5.7 is set up [Review 2nd]

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/407

Builds on:
* https://github.com/openaustralia/infrastructure/pull/408

## What does this do?

* mysql_host is now available for both mysql 5.7 and mysql 8.x servers, which greatly simplifies setup tasks
* sorted ec2-hosts in alphabetical order to make it simpler to read and compare to group_vars
* used :children in ec2-hosts to clarify when lists where a concatenation of children rather than an independent list so reasoning about group membership is simpler
* Default make now lists targets with an explanation
* Updates README to reflect actual stage names
* Adds clarifying comments ion various files
* Adds group_vars files for all groups and removes unused group to clarify what is an isnt available

## Why was this needed?

So DB setup for mysql 8 became available

And so the ansible playbooks where simpler to understand


README.md
```diff
@@ -308,7 +308,7 @@ the top of `update-ssl-certs.yaml`.
 
 You can also set:
 
-* STAGE: to a group suffix eg `STAGE=staging make apply-openaustralia` would apply changes only to openaustralia_staging
+* STAGE: to a group suffix eg `STAGE=new make apply-openaustralia` would apply changes only to `openaustralia_new`
```
benrfairless, line 311: This seems to go against our current standards - I'm not sure why we are introducing this and/or what benefit it brings?

benrfairless (changes requested):
I have reviewed this and made a number of comments. I feel like there is a lot of change here that overcomplicates things. Unless this has been previously discussed?

README.md
ianheggie-oaf: DRY is the principle. We have other projects that use the same stage= practice with capistrano. I needed to add extra targets and didn't want to have so much close but not quite the same duplication.

ianheggie-oaf:
Closed - it was puled apart into other PRs

PR #421, opened 2026-04-14
Fix/cleanup vagrantfile

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/407

## What does this do?

List ansible groups used in Vagrantfile in order, and explicitly list the normal groups as empty groups to help diagnose differences between vagrant and ec2

## Why was this needed?

Clear understanding of vagrant environment

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -62,16 +62,6 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     ansible.verbose = "vv"
 
     ansible.groups = {
-      "righttoknow" => ["righttoknow.org.au.test"],
-      "planningalerts" => ["web.planningalerts.org.au.test"],
-      "theyvoteforyou" => ["theyvoteforyou.org.au.test"],
-      "oaf" => ["oaf.org.au.test"],
-      "openaustralia" => ["openaustralia.org.au.test"],
-      "proxy" => ["au.proxy.oaf.org.au.test"],
-      "metabase" => ["web.metabase.oaf.org.au.test"],
-      "mysql" => ["mysql.test"],
-      "postgresql" => ["postgresql.test"],
-      "redis" => ["redis.test"],
       "development" => [
         "righttoknow.org.au.test",
         "web.planningalerts.org.au.test",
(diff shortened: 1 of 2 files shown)


benrfairless (commented):
Ben said:
> @ianheggie-oaf how are the servers provisioned if these aren't a member of the relevant ansible group

Everything in ec2 group gets provisioned some way - membership in extra groups adds/changes provisioning as per site.yml. Basically it:
* Installs python3, pythin3 related packages and ruby on everything
* Has two vagrant specific groups (mysql, postgresql to provide rds replacements)
* Then has all the purpose specific plays - planningalerts, theyvoteforyou, openaustralia, proxy, metabase, openvpn
* Lastly redis which replaces AWS service on vagrant

@jamezpolley ^^^

jamezpolley:
> @ianheggie-oaf how are the servers provisioned if these aren't a member of the relevant ansible group?

@benrfairless I think you're asking about the real servers. They're set up in groups in https://github.com/openaustralia/infrastructure/blob/main/inventory/ec2-hosts.

Vagrantfile
```diff
@@ -83,7 +73,29 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
+      "oaf" => ["oaf.org.au.test"],
+      "openaustralia" => ["openaustralia.org.au.test"],
+      "planningalerts" => ["web.planningalerts.org.au.test"],
+      "postgresql" => ["postgresql.test"],
+      "proxy" => ["au.proxy.oaf.org.au.test"],
+      "redis" => ["redis.test"],
+      "righttoknow" => ["righttoknow.org.au.test"],
+      "theyvoteforyou" => ["theyvoteforyou.org.au.test"],
+
+      # Server groups that are not used
```
jamezpolley: @ianheggie-oaf suggest changing comment to be more clear that these groups are used for prod, but not used in vagrant.

jamezpolley (approved):
I'm wondering if there's any value in moving these definitions into the main inventory file.

Positive - makes it more clear that things like righttoknow.org.au.test should be similar to the prod righttoknow server
Negative - running Ansible on prod is going to need a bunch of filtering to make it ignore hosts in the development group.

Okay - I think I've just talked myself out of this - makes more sense to me to keep the files separate.

Vagrantfile
ianheggie-oaf: @jamezpolley - already done - see site.yml, eg:
```
# vagrant equivalent of AWS rds_instance
- hosts: mysql
  become: true
  roles:
    - mysql
```

ianheggie-oaf: @jamezpolley - I have clarified status in comments:
```
      # Empty list just so ansible doesn't complain it doesn't know about these cloud servers
      "ec2" => [],

      # TODO: Consider adding hosts for these Server groups (that are used)
      "openvpn" => [],
      "plausible" => [],
```

Issue #205, opened 2025-08-29
Identify which services use MySQL5.7 and update to 8.4 to avoid $6.5Kpa AWS support costs now, rising to $12.7Kpa from 1/3/26


simonzippy:
From Slack...

> AWS is charging us c. USD318.72/month for "Extended Support" for our old MySQL version (5.7). If we could upgrade to v8.0 we'd go back to standard support costs until 1 August 2026; if we can upgrade to 8.4 we'd be on standard support 1 August 2029. Standard support is included in our costs - so we'd save USD318.72 every month. 5.7 moves to Year 3 pricing on 1 March 2026 - double the current cost (USD628/month) and ends entirely on 28 Feb 2027.

See the initial discussion [here >](https://oaf.slack.com/archives/C2U7XUPTR/p1756445071487279?thread_ts=1756444385.390879&cid=C2U7XUPTR) on likely systems affected and actions to take.

Br3nda:
taking this task while Simon is on leave

ianheggie-oaf:
@Br3nda (and @benrfairless ) - I have moved this to current sprint as its ongoing and I want it there to remind me the focus for this sprint / month

PR #425, opened 2026-04-15
Docs/cleanup openaustralia code to clarify

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/205

## What does this do?

makes code and project easier to understand

## Why was this needed?

Document what I had to dig to find


benrfairless (approved):
Approved. Understand there are linter failures but it's all about line length.

Issue #414, opened 2026-04-14
Fix mysql DB user and table creation for Mysql 8


ianheggie-oaf:
**Problem**

Ansible playbook doesn't handle mysql 8

Confirmed `vagrant up --provision` needs work - status as base_domain PR merged:
* https://github.com/openaustralia/infrastructure/pull/441

- [x] mysql.test - Fixed
- [ ] newprod.openaustralia.test - ok=76 FAILED=1 - openaustralia :openaustralia : reload apache - failed
- [ ] web.metabase.oaf.test - ok=67 FAILED=1 - metabase : Run metabase: Configuration error - kwargs_from_env() got an unexpected keyword argument 'ssl_version'
- [ ] web.planningalerts.test - ok=81 FAILED=1 - planningalerts : Add postgis extension to planningalerts database: Extension postgis is not available

**Describe the solution you'd like**

Set mysql_host for both DB's and avoid a lot of code duplication since the same actions are required for both 5 and 8.

**Describe alternatives you've considered**

Leaving the duplicated 5/8 code but no server needs access to both

ianheggie-oaf:
As part of this, I am running the full ansible rules on all servers on vagrant development to make sure I havn't broken other ansible roles - I am ticking off which run to the end, and if not why for us to deal with later on:

- [x] **mysql.test** *mysql : Create admin user - Access denied for user 'root'@'localhost'*
- [x] postgresql.test - *ok=52*
- [x] redis.test - *ok=47*

- [x] au.proxy.oaf.test - *ok=52*
- [ ] **newprod.openaustralia.test** -  *openaustralia : Copy SSL certificates for development failed*
- [x] oaf.test - *ok=44(
- [ ] openaustralia.test - *openaustralia : Copy SSL certificates for development failed*

---

- [ ] righttoknow.test - *zzet.rbenv : install specified ruby versions for select users* (Apply same extra package fix)
- [ ] theyvoteforyou.test - *theyvoteforyou : Apt add passenger to list - focal Release: The following signatures couldn't be verified because the public key is not available* (later, keep focused on openaustralia)
- [ ] web.metabase.oaf.test - *metabase : Run metabase - Configuration error - kwargs_from_env() got an unexpected keyword argument 'ssl_version'* (later, keep focused on openaustralia)
- [ ] web.planningalerts.test - *external/rvm.ruby : Install rubies - failed to install ruby 3.3.4* (later, keep focused on openaustralia)
[...]

PR #426, opened 2026-04-15
Fix/missing python packages for rds tasks

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/414

## What does this do?

Fixes error when ansible gathers facts from rds hosts

## Why was this needed?

Ansible error

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -33,3 +33,6 @@ github_users:
     "mlandauer",
     "simonzippy",
   ]
+
+# User python3
+ansible_python_interpreter: /usr/bin/python3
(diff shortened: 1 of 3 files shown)


group_vars/all.yml
```diff
@@ -33,3 +33,6 @@ github_users:
     "mlandauer",
     "simonzippy",
   ]
+
+# User python3
+ansible_python_interpreter: /usr/bin/python3
```
benrfairless, line 38: Should this be a group_vars or should this be in ansible.cfg?

benrfairless (approved):

group_vars/all.yml
ianheggie-oaf: @benrfairless - ansible_python_interpreter affects the target (managed) host not the local controller host, so I feel it belongs in group_vars (all.yml in this case) - I put controller config in ansible.cfg.

PR #430, opened 2026-04-16
Pin ansible-lint and yamllint to known-compatible versions

benrfairless:
## Relevant issue(s)
Nil

## What does this do?
Pins `ansible-lint` and `yamllint` to specific compatible versions in the `requirements.txt` file.

## Why was this needed?
We use old versions of Ansible. This ensures compatibility and stability of linting tools used in the project.

Users of this repo should run:
- `make clean-all`
- `make venv`
- `make roles`
- `make macos-keybase` (if required)


Makefile
```diff
@@ -182,7 +182,6 @@ update-github-ssh-keys: $(KEYSANDROLES)
 	.venv/bin/ansible-playbook site.yml --tags userkeys
 
 install-linters: venv
```
ianheggie-oaf: Given make-linters has devolved into the same as venv, lets get DRY it up and get rid of it

ianheggie-oaf (changes requested):
Lets dry this up and get rid of make-linters as it no longer serves any useful purpose

ianheggie-oaf (approved):
Approved - Looks good and makes sense to me!

PR #432, opened 2026-04-17
Update Ruby version to 4.0.2 and fix RuboCop configuration

benrfairless:
## Relevant issue(s)
- N/A

## What does this do?
- Updates Ruby version on this repository to 4.0.2 and fixes
- RuboCop configuration updates
- Rubocop linter fixes

## Why was this needed?
- This repo is using an supported version of Ruby. This updates the repository to a supported version of ruby.

- Developers should ensure they have the correct version of Ruby installed

- The infrastructure repo only uses ruby in a few files (Vagrant and rekey). This change does not flow through to servers, it is a change to the repository only .


Gemfile
```diff
@@ -2,10 +2,8 @@
 
 source "https://rubygems.org"
 
-ruby "2.7.6"
+ruby "4.0.2"
```
ianheggie-oaf, line 5: @benrfairless - You can dry up Gemfile with `ruby file: ".ruby-version"` introduced a few years ago.

.ruby-version
```diff
@@ -1 +1 @@
-2.7.6
+4.0.2
```
ianheggie-oaf, line 1: @benrfairless - I normally target one major version back from the latest to take advantage of other people testing the latest, and allowing time for gem authors to update their gems.  3.4.9 is recommended for the latest Rails, and running with rails gets a lot of attention.

ianheggie-oaf (approved):
Two non blocking recommendations:
- I suggest we aim at ruby 3.4.9 instead of the latest 4.0 because more gems will have incorporated any updates needed
- Use `ruby file: ".ruby-version"` in Gemfile to DRY it up

ianheggie-oaf:
@benrfairless - I have merged main in for you to fix the conflict in Vagrantfile.

Ok with me if you mark my comments resolved and merge - this - they where mild suggestions.

Gemfile
ianheggie-oaf: @benrfairless - I have marked this resolved, it can be picked up in a later PR

.ruby-version
ianheggie-oaf: @benrfairless - I have marked this resolved, as you pointed out elsewhere, its not on the server itself

PR #433, opened 2026-04-17
Cleanup Makefile, add checks and align with conventional goals

ianheggie-oaf:
## Relevant issue(s)

This PR made me look harder at makefile:
* https://github.com/openaustralia/infrastructure/pull/430

## What does this do?

Cleanup Makefile, add checks and align with [conventional goals](https://www.gnu.org/software/make/manual/html_node/Goals.html)

* give vagrant a chance to cleanly destroy VMs before clobbering .vagrant dir
* Rename all to help as it doesn't make anything let alone everything
* Rename everything => all as it makes everything
* Refactor macos-keybase to keybase and make it handle MacOS and Linux systems
* Add keybase to check .keybase and it's perms
* Clean whats safe to clean in terraform, echo what I am unsure of
* Rename clean-all as clobber as that is the conventional name
* Remove double spaces (cosmetic)
* Use stage convention consistently
* Adds missing dependencies
* Removes the requirement to set STAGE where there are not multiple servers
* Update README accordingly

Changes made after first review:
* Removed the requirement to set STAGE where there are not multiple servers
* Use STAGE=all when you want all stages to be selected to fix @benrfairless 's intended use
* Use .make sentinel files to stop it rerunning collections rule each time

## Why was this needed?

I felt like doing Spring cleaning in Autumn

This is donated time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,13 @@
         apply-planningalerts apply-righttoknow apply-rtk-prod apply-rtk-staging apply-theyvoteforyou \
         check-host check-metabase check-oaf check-openaustralia check-planningalerts check-righttoknow \
         check-rtk-prod check-rtk-staging check-theyvoteforyou \
-        clean clobber collections help install-linters keybase letsencrypt lint production retry roles \
+        clean clobber help install-linters keybase letsencrypt lint production retry roles \
         show-facts show-inventory show-rds-facts show-vars stage_required tf-apply tf-init tf-plan \
         update-github-ssh-keys vagrant venv yaml-lint
 
-ANSIBLE_DEPENDENCIES := .keybase roles venv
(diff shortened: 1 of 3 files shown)


Makefile
benrfairless: If I run `make check-righttoknow STAGE=` I get the following error:

` Makefile:168: *** STAGE is required, for example: STAGE=staging or STAGE=production or STAGE= for both.  Stop.`

benrfairless (approved):
Approved with one minor comment.

Makefile
ianheggie-oaf: Fixed by using STAGE=all instead of STAGE= (as otherwise the test would have been convoluted) and removing requirement for STAGE when there is only one target (as its irrelevant)

PR #437, opened 2026-04-21
Harden ImageMagick policy by restricting permissions

benrfairless:
## Relevant issue(s)
- N/A

## What does this do?
Harden the ImageMagick policy by restricting permissions for various domains.

## Why was this needed?
Enhancing security by limiting the rights associated with ImageMagick operations.

- `make apply-righttoknow STAGE=staging`
- `make apply-righttoknow STAGE=production`

- This is an urgent change required to resolve an Alaveteli security issue


roles/internal/righttoknow/tasks/main.yml
ianheggie-oaf, line 28: @benrfairless - now the ansible-lint is checking everything in roles/internal, this IS one of the checks I think we should pay attention to and add `mode: "0NNN"` to explicitly set perms.

ianheggie-oaf (changes requested):
@benrfairless I think we should set modes on files explicitly as recommended

roles/internal/righttoknow/tasks/main.yml
benrfairless: Agree but this should be in its own PR. I'll create one.

benrfairless:
> @benrfairless I think we should set modes on files explicitly as recommended

I agree @ianheggie-oaf - but that isn't a place for this PR. I'll work on one tomorrow to handle that.

ianheggie-oaf:
> > @benrfairless I think we should set modes on files explicitly as recommended
> 
> I agree @ianheggie-oaf - but that isn't a place for this PR. I'll work on one tomorrow to handle that.

@benrfairless - Makes sense.

ianheggie-oaf (approved):
Approved - agreed with ben for File perms to be a separate PR

PR #439, opened 2026-04-21
Add HelpScout DKIM records for multiple domains

benrfairless:
## What does this do?
Adds DKIM records for HelpScout across multiple domains in the DNS configuration.

## Why was this needed?
We are implementing HelpScout and this is a requirement for emails

- Run `make tf-plan` to verify
- Run `make tf-apply` to apply changes

- I have verified on Helpscout that a SPF record is _not_ required.


ianheggie-oaf (approved):
It makes sense at a glance and it passes format and validation checks, based on:
* https://github.com/openaustralia/infrastructure/pull/446

(I had AWS but not the other credentials so plan only partially ran before failing, and didn’t get to the Cloudflare part of the plan)

Issue #438, opened 2026-04-21
"Failed to update apt cache" when running TASK [Install python packages]


benrfairless:
**Describe the bug**
When running `make apply-righttoknow STAGE=all` it fails to run intermittently with the error `failed: [staging.righttoknow.org.au] (item=python3-pip) => {"ansible_loop_var": "item", "changed": false, "item": "python3-pip", "msg": "Failed to update apt cache: unknown reason"}`

**To Reproduce**
1. Run `.venv/bin/ansible-playbook  -i ./inventory/ec2-hosts site.yml -l righttoknow --diff`
2. Look out for the `TASK [Install python packages]` task

**Expected behavior**
The package should either be installed or not

**Screenshots/Output**

Output below:

```
TASK [Install python packages] ********************************************************************************************************************************************************
failed: [prod.righttoknow.org.au] (item=python3-pip) => {"ansible_loop_var": "item", "changed": false, "item": "python3-pip", "msg": "Failed to update apt cache: unknown reason"}
ok: [prod.righttoknow.org.au] => (item=python3-pymysql)
failed: [staging.righttoknow.org.au] (item=python3-pip) => {"ansible_loop_var": "item", "changed": false, "item": "python3-pip", "msg": "Failed to update apt cache: unknown reason"}
ok: [prod.righttoknow.org.au] => (item=python3-psycopg2)
ok: [staging.righttoknow.org.au] => (item=python3-pymysql)
ok: [staging.righttoknow.org.au] => (item=python3-psycopg2)
```

**Additional context**
This is intermittent, and the task seems to run fine if you re-run the command shortly after it fails

ianheggie-oaf:
@benrfairless - I have come across this if unattended-updates package is installed.

I have a suggested PR when we upgrade to supported ansible version (> 2.12):
* https://github.com/openaustralia/infrastructure/pull/440


- I just checked and the ansible rules do add unattended-updates:
```
================== mysql.test ===================
ii  unattended-upgrades              1.1ubuntu1.18.04.7~16.04.7                 all          automatic installation of security upgrades
================== postgresql.test ===================
ii  unattended-upgrades              2.8ubuntu1                              all          automatic installation of security upgrades
================== au.proxy.oaf.test ===================
_____________________________________________________________________
ii  unattended-upgrades              1.1ubuntu1.18.04.7~16.04.7                 all          automatic installation of security upgrades
================== redis.test ===================
[...]

PR #440, opened 2026-04-21
Increase apt lock timeout from 60 second default to 300 (upgrade of ansible required)

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/438

NOTE: We will have to upgrade from ansible 2.10 (EOL May 2022) to 2.18 (released in late 2024 and matches the python v3.11.13 we are using)

## What does this do?

Fixes lock conflict if unattended-updates is running in background for internal roles

Increase apt lock timeout from 60 second default to 300

Note - external roles may still fail

## Why was this needed?

Ben said:
> When running make apply-righttoknow STAGE=all it fails to run intermittently with the error failed: [staging.righttoknow.org.au] (item=python3-pip) => {"ansible_loop_var": "item", "changed": false, "item": "python3-pip", "msg": "Failed to update apt cache: unknown reason"}

I have encountered this issue on other projects.

We can up the timeout further if needed - the first time it runs on a new VM it will take time.

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -40,3 +40,6 @@ github_users:
 
 # User python3
 ansible_python_interpreter: /usr/bin/python3
+
+# Fix "Failed to update apt cache" errors by increasing timeout from default 60 seconds
+apt_lock_timeout: 300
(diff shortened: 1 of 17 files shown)


ianheggie-oaf:
@benrfairless - are we likely to upgrade ansible? if so this is applicable and can come off tcebox

Issue #424, opened 2026-04-15
Make sure newprod and staging emails go to mailpit not real people


ianheggie-oaf:
We don't want changes in hansard to trigger emails to real people from newprod or staging

ianheggie-oaf:
@Br3nda - make into an independent issue

ianheggie-oaf:
Closed as the existing system handles development, and staging is not a copy of production

PR #456, opened 2026-04-26
Feature catch emails for staging and non live servers (deferred)

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/424

## What does this do?

Configures staging and non live servers to catch email

## Why was this needed?

unable to properly test server before going live without it as it turned out there needed to be a multi faceted solution to handle both apache vhost and cron jobs


Br3nda:
I don't think we need this anymore once the server is live

ianheggie-oaf:
> I don't think we need this anymore once the server is live

@Br3nda - still useful for test (staging env on server that has live production)

ianheggie-oaf:
@Br3nda - Im thinking this has become irrelevant - should I close it without merging?

ianheggie-oaf:
Abandoned, staging apparently shouldn’t have a copy of production data

PR #459, opened 2026-04-27
Fix/go live

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/205

## What does this do?

Fix what doesn’t work now we are merging to main and going live

Its effectively already merged since PR https://github.com/openaustralia/infrastructure/pull/463
was accidentality merged by GitHub (rebasing when auto mergfe is on makes it merge)


benrfairless (approved):
Approving as @ianheggie-oaf has said that "its been used to provision openaustralia anyway"

PR #461, opened 2026-04-27
Feature/disable cron

ianheggie-oaf:
## Relevant issue(s)

Slack:
Brenda  [6:14 PM]
[@Ian H](https://oaf.slack.com/team/U07QEL1PF6J) can you disable the cron? I need to test ks before it runs and my laptop is borked

## What does this do?

Builds on:

* https://github.com/openaustralia/infrastructure/pull/459

Add CRON_DISABLED ENV var to disable cron

## Why was this needed?

Brenda requested, and I hate "its only temporary" code changes with a side dish of "do this again after ansible is run"


ianheggie-oaf:
Decided that disabling cron needs to be a committed change not and ENV variable

Issue #471, opened 2026-04-28
Fix xapian and SSL ansible rules


ianheggie-oaf:
**Describe the bug**

Ansible role fails to compile xapian:
```
php/xapian_wrap.cc:44649:122: error: 'SNIPPET_NGRAMS' is not a member of 'Xapian::MSet'
FLAG_NGRAMS and FLAG_NO_POSITIONS are also missing
```

NOTE: newprod was manually made with a mixture of versions:

* libxapian-dev                      1.4.22-1.999
* libsearch-xapian-perl              1.2.25.4-1build1
* xapian-bindings-1.4.25.tar.xz  
*  xapian-core-1.4.25.tar.xz

**To Reproduce**
1. Provision a fresh vagrant server using the Ansible role: `make openaustralia.test`
2. See it fail at "openaustralia : Compile xapian bindings"

**Expected behavior**

`vagrant up --provision openaustralia.test` runs successfully

**Additional context**

You need to run: `vagrant up mysql.test` first

ianheggie-oaf:
@benrfairless or @Br3nda - moved to current sprint, waiting on PRs to be reviewed and merged to avoid the danger of previous versions of playbook being actioned by mistake

Issue #420, opened 2026-04-14
Investigate why newprod server recently frooze twice in a row and mitigate if possible


ianheggie-oaf:
**Describe the bug**

Wen to work on staging server and got:

ssh: connect to host newprod.openaustralia.org.au port 22: Connection timed out

**To Reproduce**

```
$->(2) make show-facts host=newprod.openaustralia.org.au
.venv/bin/ansible -i ./inventory/ec2-hosts newprod.openaustralia.org.au -m setup
newprod.openaustralia.org.au | UNREACHABLE! => {
    "changed": false,
    "msg": "Failed to connect to the host via ssh: ssh: connect to host newprod.openaustralia.org.au port 22: Connection timed out",
    "unreachable": true
}
```

**Expected behavior**

It should connect via ssh and list facts

**Screenshots/Output**

```
$->(130) ssh -v oaoas
OpenSSH_9.6p1 Ubuntu-3ubuntu13.15, OpenSSL 3.0.13 30 Jan 2024
debug1: Reading configuration data /home/ianh/.ssh/config
debug1: /home/ianh/.ssh/config line 8: Applying options for *
debug1: /home/ianh/.ssh/config line 110: Applying options for oaoas
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 19: include /etc/ssh/ssh_config.d/*.conf matched no files
debug1: /etc/ssh/ssh_config line 21: Applying options for *
debug1: Connecting to 3.24.248.95 [3.24.248.95] port 22.
```

**Additional context**

I have added a password so I can check what is happening over the serial console next time.

ianheggie-oaf:
As discussed with @benrfairless and @Br3nda - I (Ian) will 
1. pull down a copy of production database and profile cron jobs to see how much memory they take. 
2. Look at the process stats and estimate memory usage of processes we don't use
3. Add 1/2 to 1 GB swap space to openaustralia server to give a bit more lee way before server becomes unresponsive (Basically estimating non working set memory and adding enough swap to allow the working set to use the full 2GB of memory and push whats not being used to swap)

ianheggie-oaf:
@benrfairless and @Br3nda - Looking at the existing server, with
```
zgrep -i "oom\|killed process\|out of memory" /var/log/syslog* | grep -v "^Binary"

[...]

PR #472, opened 2026-04-28
fix certificate generation for development

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/471 - added for this problem
* https://github.com/openaustralia/infrastructure/issues/420 - discovered the problem whilst researching this

## What does this do?

* Handles blog being created with blog.openaustralia.org for production and using base_domain for development
* Create missing ext files for domains not previously used in development
* Adds `make generate-certificates` target as per `make help`
* Updates ext files to use .test rather than .org.au.test fro development domain suffixes

## Why was this needed?

provisioning openaustralia.test failed when trying to install SSL certificates,
which was needed for 

```
make vagrant
vagrant up mysql.test
vagrant up --provision openaustralia.test
```


ianheggie-oaf:
I have enabled auto merge on the basis that I have tested it on my own dev system and it works.

benrfairless (commented):
I'm happy with these changes, however can you confirm there is no documentation that needs to be updated? If that's all fine then please re-submit for approval.

ianheggie-oaf:
> I'm happy with these changes, however can you confirm there is no documentation that needs to be updated? If that's all fine then please re-submit for approval.

@benrfairless I don't believe it needs a change in documentation because:
* It provides the make target for `make generate-certificates` that is already documented in `make help`
* It fixes certificates/generate-certificates.sh called by that target (broken when I changed development domain names as already documented in README.md)

Please review

benrfairless (approved):

PR #473, opened 2026-04-28
Fix/xapian ansible rules

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/471
* https://github.com/openaustralia/infrastructure/issues/420 - discovered the problem whilst researching this

## What does this do?

Builds on the following PR - review and merge it FIRST:
* https://github.com/openaustralia/infrastructure/pull/472

* Always installs xapian for openaustralia
* Adds xapian package source as manually configured on newprod
* Auto detects `xapian_version` from the libxapian-dev package that was installed (removed the option to disable xapian build)
* Fixes xapian being built with a different version of the xapian library
* Adds xapian tags to make it quick to rerun the provisioning

```
TAGS=xapian time vagrant up --provision openaustralia.test
``` 

## Why was this needed?

I was unable to provision openaustralia.test.
newprod only worked because the ansible rules skip when an existing installation is detected

## Notes to reviewer (Optional)

remember to `make vagrant` and `vagrant up mysql.test ` first

PR #474, opened 2026-04-28
Feature/ add swap space

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/infrastructure/issues/420

## What does this do?

Adds 1GB of swap space so I can discover how much memory the xapian index process actually needs

## Why was this needed?

OOM-kill has killed a number of perl script cron jobs in the last month at the 1,600 MB mark. Adding swap hopefully allows it to run to completion so we can get statistics.

Tested on top of these PRs so `vagrant up --provision openaustralia.test` works:

* https://github.com/openaustralia/infrastructure/pull/472
* https://github.com/openaustralia/infrastructure/pull/473

( i have a prs local branch that merges all of these PRs)

I have applied this change to the live server:

```
TAGS=swap make check-openaustralia
TAGS=swap make apply-openaustralia
```

### DIFF
--- a/group_vars/openaustralia.yml
+++ b/group_vars/openaustralia.yml
@@ -171,3 +171,20 @@ apache_sites:
   software:
     domain: "software.{{ openaustralia_domain }}"
     server_aliases: "{{ (base_domain == 'org.au') | ternary(extra_software_server_aliases, []) }}"
+
+
+# Sized to cover (approximate figures):
+#  1,600 MB Xapian anon (index.pl cron job)
+#           -> true peak will be higher, only observable once swap is added
+#    120 MB Idle processes (snapd, cloudwatch, newrelic, postfix etc)
+#           -> can swap with negligible impact
+#    168 MB Apache workers anon
+#           -> can be swapped but with site performance impact
+#     42 MB Locked in memory (unevictable kernel pages, mlocked)
+#  --------
+#  1,930 MB Physical RAM
+
+# If we need more than this, we need to upgrade the ec2 instance size!
+swap_file_size_mb: '1024'
+# Prefer RAM but swap out genuinely idle processes under pressure
+swap_swappiness: '35'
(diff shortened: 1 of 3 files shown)


group_vars/openaustralia.yml
```diff
@@ -170,4 +170,21 @@ apache_sites:
...
+#    168 MB Apache workers anon
+#           -> can be swapped but with site performance impact
+#     42 MB Locked in memory (unevictable kernel pages, mlocked)
+#  --------
+#  1,930 MB Physical RAM
+
+# If we need more than this, we need to upgrade the ec2 instance size!
+swap_file_size_mb: '1024'
+# Prefer RAM but swap out genuinely idle processes under pressure
+swap_swappiness: '35'
```
benrfairless, line 190: I'll approve this, but the name `swappiness` doesn't really describe what it's doing....

benrfairless (approved):

group_vars/openaustralia.yml
ianheggie-oaf: @benrfairless - I am curious what name you would suggest instead? though after 2 decades of that name being in the linux kernel, It would be "rather hard" to  convince Linus to change it! Jeff Geerling (the author of the ansible role I used) chose to keep with the name of the sysctl param, and I think he made the right call.

I looked at
* https://docs.kernel.org/admin-guide/sysctl/vm.html
* https://kubernetes.io/blog/2025/08/19/tuning-linux-swap-for-kubernetes-a-deep-dive/
* and looked at the archives for the linux kernel code (this predates the git repo I checked first) - interesting how it has a "distress" value of how much trouble its having reclaiming pages, and swap_tendency derived from a mapping ratio, distress and swappiness - it feels like when you really get down into the mm/vmscan.c code I found in linux-2.5.75 (near when it was introduced) it looks like there was a lot of fudging and best guessing.

PR #488, opened 2026-05-05
Speed up ansible mysql provisioning by caching apt update for 1 hour

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/openaustralia/issues/788

## What does this do?

Speeds up ansible mysql provisioning by caching apt update for 1 hour (mysqld server for vagrant)

## Why was this needed?

Found inefficiency when I was working on openaustralia rules and I applied it elsewhere for consistancy

The speed up is when apt is refreshed multiple times

### DIFF
--- a/roles/internal/mysql/tasks/main.yml
+++ b/roles/internal/mysql/tasks/main.yml
@@ -6,6 +6,7 @@
   apt:
     pkg: mysql-server
     update_cache: true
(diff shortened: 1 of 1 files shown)


benrfairless (approved):
Love this! We should probably expand this so that we only pull the apt cache once every 5mins for every task.

PR #498, opened 2026-05-12
Fix apache log path to end in .log and correct dir

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/openaustralia/issues/788

## What does this do?

Updates ansible rules to make all log files end in _log rather than .log so logrotate manages them.

## Why was this needed?

inconsistent log suffix

These rules have already been applied (for _log ending) using

```
TAGS=sites_available make check-openaustralia
TAGS=sites_available make apply-openaustralia
```

(Depends on another PR to add the tags)

### DIFF
--- a/group_vars/openaustralia.yml
+++ b/group_vars/openaustralia.yml
@@ -72,8 +72,8 @@ logs:
      log_stream_name: "{{ inventory_hostname }}"
      timestamp_format: "%b %d %H:%M:%S"
 
-   - file_path: /srv/www/production/log/error_log
-     log_group_name: /srv/www/production/log/error_log
+   - file_path: /srv/www/production/log/error.log
+     log_group_name: /srv/www/production/log/error.log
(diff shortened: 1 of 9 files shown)


benrfairless (changes requested):
@ianheggie-oaf we have logrotate configured on other services (such as righttoknow) and they work file with the logfile ending with .log

I would like there to be some sort of consistency with how log files are managed across our services if possible.

ianheggie-oaf:
> @ianheggie-oaf we have logrotate configured on other services (such as righttoknow) and they work file with the logfile ending with .log
> 
> I would like there to be some sort of consistency with how log files are managed across our services if possible.

@benrfairless - I AGREE - I wasn't sure why _log was used on this server, I will update this PR accordingly

ianheggie-oaf:
@benrfairless updated to use .log - pls review

benrfairless (approved):

PR #501, opened 2026-05-13
Update database host references to db_host

benrfairless:
## What does this do?
Updates the remaining references of `postgresql_host` to `db_host` in configuration files and tasks.

## Why was this needed?
We changed the variable from postgresql_host to db_host

No deployment required (however you can run `make check` commands


ianheggie-oaf (approved):
Makes sense to me - approved

PR #515, opened 2026-05-14
Add non-proxy A record for srv.theyvoteforyou.org.au

benrfairless:
## What does this do?
Adds a non-proxy A record for the domain srv.theyvoteforyou.org.au.

## Why was this needed?
This change allows direct access to the server without proxying through Cloudflare, which is necessary until OpenVPN is configured.

Do not deploy this - change has been made manually - there is a config misalignment in terraform currently.

Changes have been applied manually in Cloudflare

### DIFF
--- a/terraform/theyvoteforyou/dns.tf
+++ b/terraform/theyvoteforyou/dns.tf
@@ -27,6 +27,15 @@ resource "cloudflare_record" "root" {
   proxied = true
 }
 
+### TODO: Remove this when we have the OpenVPN configured
+resource "cloudflare_record" "non_proxy_root" {
+  zone_id = cloudflare_zone.org_au.id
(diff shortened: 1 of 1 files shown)


ianheggie-oaf (approved):
Approved - makes sense to me.

BTW - I am thinking instead of removing it when OpenVPN is configured, making the servers:
1. have http user/password blocking direct access (so it stops everyone else but us and our linkcheckers)
2. Either configure the server to use cloudflaretunnel or to allow the cloudflare egress (proxy source) IP list (and have it check each 3 hours and update list as needed because the IP list can change with roughly 4-6 hours minimum between change and usage) 

The alternative is we require developers to use ssh or openvpn.

What do you think @benrfairless ?

PR #517, opened 2026-05-15
Update Terraform configuration to reflect current state

benrfairless:
## Relevant issue(s)
- https://github.com/openaustralia/infrastructure/issues/516

## What does this do?
1. Comments out the MySQL 5.6 database so that terraform doesn't try to recreate it
2. Renames all the old openaustralia.org AWS config to OLD so we can see that in AWS

## Why was this needed?
1. We don't want to create a new MySQL 5.6 DB by accident
2. We should be able to tell the difference between old and new servers in AWS

1. `make tf-plan` to verify
2. `make tf-apply` to apply changes

There are a number of DNS changes which were approved in a previous PR - these are additions only and are safe to apply.


ianheggie-oaf (approved):
Approved - makes sense to me

PR #519, opened 2026-05-21
Run tvfy test site with RAILS_ENV=staging 

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/theyvoteforyou/issues/1612

## What does this do?

* Sets RAILS_ENV=staging when running TVFY staging test web site
* Adds production ip filter to staging to keep them as close as possible

## Why was this needed?

Rails app was misconfigured as production for the test web site


roles/internal/theyvoteforyou/templates/staging
```diff
@@ -20,6 +22,577 @@ server {
...
+    deny 94.244.144.0/20;
+    deny 89.150.208.0/20;
+    deny 148.145.224.0/23;
+    deny 83.101.112.0/21;
+    deny 80.238.191.0/24;
+    deny 2405:f080:3401::/48;
+    deny 110.41.90.0/24;
+    deny 83.101.120.0/21;
+  }
+
```
benrfairless, line 595: Why does this exist? We should be using Cloudflare to mitigate scraping/DDOS errors?

benrfairless (changes requested):

roles/internal/theyvoteforyou/templates/staging
ianheggie-oaf: > Why does this exist?

@benrfairless - This is to make staging as like production as possible - test things in staging first, then apply to production when automated tests can't check it.

> We should be using Cloudflare to mitigate scraping/DDOS errors?

i agree, and when we do the following both staging and production should be updated accordingly:
* https://github.com/openaustralia/infrastructure/issues/312

benrfairless (approved):

PR #521, opened 2026-05-25
Update GitHub user management for deployment

benrfairless:
## Relevant issue(s)
- https://github.com/openaustralia/oaf-internal/issues/231

## What does this do?
- Removes the specific github_users list for righttoknow
- Remove jattree, mlandauer, simonzippy from server access
- Update deploy-user ansible role to support the removal of users/keys.
- Add safety gate to deploy-user ansible role so that it doesn't manage authorised keys without a github_users list
- Ensure that terraform.pem is included in the authorised keys list.

## Why was this needed?
- Makes it easier to add/remove users to deploy servers
- Removes users who no longer need SSH access to servers
- Ensures that only the correct keys are configured on each server

- Run `make update-github-ssh-keys` to apply to all servers

- Please double check this configuration before approving, as it is destructive (eg: removes keys).

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -37,9 +37,6 @@ github_users:
     "br3nda",
     "ianheggie-oaf",
     "jamezpolley",
-    "jattree",
-    "mlandauer",
-    "simonzippy",
   ]
 
 
--- a/group_vars/righttoknow.yml
+++ b/group_vars/righttoknow.yml
@@ -1,16 +1,7 @@
 # TODO: Unify this with the domain. It's currently only used to name
 # things on New Relic.
 site_name: righttoknow.org.au
-github_users:
-  [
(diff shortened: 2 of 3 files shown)


jamezpolley (approved):

ianheggie-oaf (approved):
Approved - I like it cleans out old keys, and is consistent. Only think left is a discussion if we make it ubuntu only, no root access except on console - root gets a lot of door knocking, but then so does the ssh port itself. Anyway, a discussion for another time.

benrfairless:
> Approved - I like it cleans out old keys, and is consistent. Only think left is a discussion if we make it ubuntu only, no root access except on console - root gets a lot of door knocking, but then so does the ssh port itself. Anyway, a discussion for another time.

I'm going to address this in another PR 😉

PR #525, opened 2026-05-25
Update name of storage.yml file for Right to Know ansible role

benrfairless:
## What does this do?
Renames the `storage.yml` file so that ansible can process it.

## Why was this needed?
The file hasn't changed, but ansible needs the file to end with `.yml` and not `.yml.j2`.

No implementation/deployment steps required


ianheggie-oaf (changes requested):
@benrfairless I recommend that all templates end in .j2 as they ARE j2 templates first, and then the second extension second, and should be validated as such, it is/should be best practice:
 https://docs.ansible.com/projects/lint/rules/jinja-template-extension/

If a file doesn't have any j2 expressions, then it shouldn’t be a template, instead place under files, use copy not template command, and in that case no .j2 extension.

BTW - its the same convention that something.html.erb follows - its an erb file that produces a html file. In this case its a j2 file that produces a yml file.

benrfairless (commented):
Made the following updates:
- Tagged the alaveteli config files as "alaveteli"
- Removed a duplicate storage.yml task
- Moved the storage.yml file from templates to tasks

Test Results (truncated):
```
❯ make check-righttoknow TAGS=alaveteli STAGE=all
...
PLAY [righttoknow] ******************************************************************************************************************************************************************************************************

TASK [righttoknow : Copy over database configuration for application] ***************************************************************************************************************************************************
ok: [prod.righttoknow.org.au] => (item=production)
ok: [staging.righttoknow.org.au] => (item=staging)

TASK [righttoknow : Apply Alaveteli config (production)] ****************************************************************************************************************************************************************
skipping: [staging.righttoknow.org.au]
ok: [prod.righttoknow.org.au]

TASK [righttoknow : Apply Alaveteli config (staging)] *******************************************************************************************************************************************************************
skipping: [prod.righttoknow.org.au]
ok: [staging.righttoknow.org.au]

TASK [righttoknow : Apply Alaveteli Storage Config] *********************************************************************************************************************************************************************
ok: [prod.righttoknow.org.au] => (item=production)
--- before
+++ after
@@ -1,4 +1,4 @@
 {
-    "mode": "0664",
+    "mode": "0644",
     "path": "/srv/www/staging/shared/storage.yml"
 }

changed: [staging.righttoknow.org.au] => (item=staging)

TASK [righttoknow : Apply user spam scorer config] **********************************************************************************************************************************************************************
ok: [staging.righttoknow.org.au] => (item=staging)
ok: [prod.righttoknow.org.au] => (item=production)

TASK [righttoknow : Apply sidekiq config] *******************************************************************************************************************************************************************************
ok: [staging.righttoknow.org.au] => (item=staging)
ok: [prod.righttoknow.org.au] => (item=production)

RUNNING HANDLER [righttoknow : nginx restart] ***************************************************************************************************************************************************************************
changed: [staging.righttoknow.org.au]
```

benrfairless (commented):
Linters are failing because it doesn't like the actual storage.yml file - which is currently in use on the server and shouldn't be modified.

Br3nda:
some linting checks still to fix (i came by because the slack bot reminded me)

Br3nda (approved):

Br3nda:
saw your comment.. totally :+1: LGTM

Issue #490, opened 2026-05-07
Document how to use 1Password in lieu of Keybase


benrfairless:
**Is your feature request related to a problem? Please describe.**
We currently use Keybase which is a headache for secrets. We now have 1Password which could be a much better option!

**Describe the solution you'd like**
@jamezpolley mentioned he was using 1Password instead of Keybase. This should be documented (possibly in https://github.com/openaustralia/oaf-internal)

**Describe alternatives you've considered**
[fnox](https://github.com/jdx/fnox) might be a solution we could look at down the track - but if James has 1Password working already that would be good enough.

**Additional context**
If my understanding here is wrong please let me know and we can revisit.

ianheggie-oaf:
Testing steps listed in following PR as requested:
* https://github.com/openaustralia/infrastructure/pull/531

Issue #238, opened 2025-10-09
[BUG] Keybase Integration broken on MacOS


benrfairless:
**Describe the bug**
Running `make` does not correctly symlink the keybase folder on MacOS.

**To Reproduce**
1. Checkout main branch of infrastructure
2. run make

**Expected behavior**
a .keybase folder should be created that points to the correct location in Keybase

**Screenshots/Output**
Error message on MacOS: 
```
❯  keybase config get -d -b mountdir 
▶ ERROR <root>.mountdir: no such key: mountdir```
```

**Additional context**
Caused by #231, requesting this be reverted for the time being.

PR #531, opened 2026-05-26
Move secrets distribution from Keybase to 1Password

benrfairless:
## Relevant issue(s)
- #490 
- #238 

## What does this do?

Moves the five Keybase-distributed files this repo depends on (four `ansible-vault` passphrases + the `terraform.pem` SSH key) so they come from the OAF 1Password account instead. **1Password is now the sole source** — the Keybase runtime fallback has been removed, so being signed into the OAF 1Password account is required.

Concretely:

- `bin/ansible-vault-client` is a `--client`-mode dispatcher. `ansible.cfg`'s `vault_identity_list` now points every vault id at it. The dispatcher reads from the OAF 1Password account (pinning the account and using vault/item UUIDs, so a contributor signed into multiple 1Password accounts can't accidentally fetch from the wrong one). If `op` can't read the passphrase it fails cleanly with a "sign into 1Password" message — there is no Keybase fallback.
- `bin/fetch-terraform-pem` writes `terraform.pem` from 1Password to a gitignored regular file. It's idempotent (no-op if the file is already present) and requires the 1Password CLI. The previously tracked Keybase symlink has been removed.
- Terraform: the per-operator AWS access/secret-key variables are gone — the AWS provider now uses each operator's AWS credential chain, and Google uses gcloud ADC. The shared secrets — `rds_admin_password`, `cloudflare_api_token`, and `linode_api_token` — are rendered into `terraform/secrets.auto.tfvars` from the OAF DevOps 1Password vault by `make tf-secrets` (via `op inject`). The dead `theyvoteforyou_db_password` variable (declared but never referenced) is removed.
- New `Makefile` targets: `op-check` (fatal — fails fast with a clear error if you aren't signed into the OAF 1Password account), `terraform.pem` (fetch on demand), `tf-secrets` (render tfvars), `tf-env-check` (warn about missing per-user creds). `requirements` depends on `op-check terraform.pem` instead of the Keybase mount.
- The Keybase machinery is fully removed: the `keybase` / `.keybase` Makefile targets, `bin/headless-keybase.sh`, and the four committed `.*-vault-pass` symlinks are deleted.
- `packer/planningalerts.pkr.hcl` no longer passes the `.*-vault-pass` symlinks as `--vault-id`; it points at the `bin/ansible-vault-client` dispatcher instead, matching `ansible.cfg`. (The Packer build host therefore needs to be signed into the OAF 1Password account.)
- `bin/rotate-vault-passphrase <id>` walks `group_vars/` and `host_vars/`, re-encrypting every `!vault` block tagged with that id and writing the new passphrase to 1Password. The flow is defensive (1Password first, files second) with a state-aware EXIT trap that preserves OLD + NEW passphrases and prints copy-pasteable recovery instructions if the file rewrite fails after 1Password was updated. The new passphrase never appears in `argv` — it goes through `jq --rawfile` and `op item edit`'s stdin template. Handles both 1.2-tagged and 1.1-unlabeled blocks for the `default` id.
- `bin/rekey-yaml` is the engine. It replaces the old `rekey.rb` (which used Psych and destroyed comments/key order) and works in text mode so only `!vault` blocks change. `--in-place` writes go through a temp file + atomic rename. Errors carry the file path and block header so rotation failures are self-diagnosing.
- README rewrites cover the 1Password workflow, how Terraform sources its credentials (AWS/gcloud from your own CLI tooling; the shared RDS/Cloudflare/Linode secrets from 1Password via `make tf-secrets`), and how to rotate a vault passphrase.

## Why was this needed?

Keybase is no longer actively developed, the desktop client has been drifting on macOS and Linux, and getting a new joiner set up has gotten progressively more painful. We already use 1Password elsewhere (commits are SSH-signed via it), so consolidating onto a single secrets tool removes a moving part and the 'Keybase isn't mounting, can someone help' Slack thread.

Already done as part of this PR:
- 1Password items live in the OAF `DevOps` and `RTK Devops` vaults — UUIDs are baked into the dispatcher and templates. Passphrase values were verified byte-for-byte against Keybase before commit.
- `terraform validate` passes (no straggling refs to the removed variables in any submodule).
- `make ansible-lint`, `make tf-check-fmt` still green.
- Decrypt round-trip verified end-to-end.
- Rotation script dry-run for every vault id passes; `default` (the worst case at 7 files) finishes in ~20 s.

What a reviewer/merger needs to do:
1. Make sure every current operator has been added to the OAF `DevOps` vault (and `RTK Devops` for Right To Know admins) **before** merging — 1Password is now required, there is no Keybase fallback.
2. Merge and tell the team: 'Run `make clean && make requirements`. You must be signed into the OAF 1Password account (`op signin --account oaforgau`); `make` will fail fast at `op-check` otherwise.' Anyone with a stray `.keybase` symlink can delete it — nothing uses it any more.

- The dispatcher is invoked by every `ansible-vault` subprocess. If you `op signout`, the script now fails cleanly (it no longer falls back to Keybase) — verify by `op signout && bin/ansible-vault-client --vault-id ec2` (it should print a "sign into 1Password" error and exit non-zero).
- Per-operator credentials (AWS, gcloud) come from each operator's own CLI tooling rather than 1Password, so nobody is forced to keep their personal keys in 1Password. The Cloudflare and Linode tokens are the exception: they're shared service tokens, so they live in the OAF DevOps 1Password vault and are rendered into `secrets.auto.tfvars` by `make tf-secrets`.
- I did **not** rotate any of the four vault passphrases as part of this PR — the values in 1Password are byte-identical to the ones that were in Keybase. The `bin/rotate-vault-passphrase` script exists so future rotation is a single command, but it's untested against a live rotation. Recommend dry-running it (`bin/rotate-vault-passphrase ec2 --dry-run`) before the first real use, ideally on a feature branch so the diff is easy to revert.
- `op item edit` exposes argv to other local processes via `ps`; the rotation script avoids that by piping a JSON template via stdin (the only way `op` recommends for sensitive values).


benrfairless:
@ianheggie-oaf @Br3nda @ianheggie-oaf I have tested this and it works for me - I would like you to make sure it works for you as well. Can you test and let me know?

Br3nda:
i haven't tested yet! but i will this weekend

bin/ansible-vault-client
```diff
@@ -0,0 +1,52 @@
...
+    --vault-id=*) vault_id="${1#*=}"; shift ;;
+    *) shift ;;
+  esac
+done
+
+case "$vault_id" in
+  default) op_ref="op://nal3r3nlyo2sdqzegvdvlmvp5e/m2ccf6rvruxlyhevcovwher4cy/password"; kb=".vault_pass.txt" ;;
+  all)     op_ref="op://nal3r3nlyo2sdqzegvdvlmvp5e/mpwsqfdmg7xinu46zmiqesn2yu/password"; kb=".all-vault-pass" ;;
+  ec2)     op_ref="op://nal3r3nlyo2sdqzegvdvlmvp5e/emvvhq7denilwx2nesg45fkwrm/password"; kb=".ec2-vault-pass" ;;
+  rtk)     op_ref="op://nal3r3nlyo2sdqzegvdvlmvp5e/pifybc4cdnm7ah2sjrgcpgibiy/password"; kb=".rtk-vault-pass" ;;
```
ianheggie-oaf:
```suggestion
  rtk)     op_ref="op://nal3r3nlyo2sdqzegvdvlmvp5e/33howvcuao7gomvqpivs7dtzoy/password"; kb=".rtk-vault-pass" ;;
```

ianheggie-oaf: Retrieve from "Password" called ".rtk-vault-pass" since the "API Credentials" with the same name does not have a `password` attribute.
Alternatively you could have used
op_ref="op://nal3r3nlyo2sdqzegvdvlmvp5e/pifybc4cdnm7ah2sjrgcpgibiy/credential" 
which also works, but is inconsistent with the other passwords.

I suggest rtk's secrets is moved to a Password or Document entry  and the duplicated API credentials removed in 1password.

terraform/providers.tf
```diff
@@ -1,7 +1,11 @@
+# Provider credentials come from each operator's CLI tooling, not tfvars:
+#   AWS         — `~/.aws/credentials`, AWS_PROFILE, AWS SSO, instance profile
```
ianheggie-oaf:
```suggestion
#   AWS         — `~/.aws/credentials`, AWS_PROFILE, AWS SSO, instance profile, or AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY env vars
```

ianheggie-oaf: Add the aws config vars I use in .envrc

ianheggie-oaf:
Testing results:

* bin/ansible-vault-client has a bug - I suggested change fixes it for me (testing without fallback .keybase reveals the bug)

* I did not test
   bin/rekey-yaml,
   bin/rotate-vault-passphrase
   
* I tested the other new make goals, and tf-init and tf-plan - they work (given the suggested change above with 1password and no keybase)

ianheggie-oaf (changes requested):
Requires a change to get it working for me when .keybase is not present

benrfairless:
@ianheggie-oaf I've made a few changes to this script:
1. Implemented the recommendations you have provided
2. Created Account level Cloudflare and Linode API keys, saved these in 1Password, and inject them automatically when creating `secrets.tfvars`

@jamezpolley @ianheggie-oaf can you review these changes please?

ianheggie-oaf (dismissed):
1. Remove requirement for "Keybase still works as a runtime fallback" or fix. Though once `make check-planningalerts` has been run with sign in to 1password, you can sign out and it still works using .keybase, so its not a big issue - at least up to another problem i have raised the following issue for:
  * https://github.com/openaustralia/infrastructure/issues/546 make check-planningalerts fails with 'newrelic_license_key' is undefined

RE:

* terraform validate passes

Passed - checked using `make clean` then `make tf-validate` for both keybase only and 1password only

* make ansible-lint
 
Passed without keybase or 1password
 
* make tf-check-fmt

Passed without keybase or 1password

* Decrypt round-trip verified end-to-end via ansible -m debug -a 'var=rds_admin_password'.

This fails for me - I used instead:
```
make show-vars host=ec2-3-27-189-101.ap-southeast-2.compute.amazonaws.com | egrep password
make show-vars host=theyvoteforyou.org.au | egrep password
make show-vars host=openaustralia.org.au | egrep password
```

* Make sure every current operator has been added to the OAF DevOps vault.

Checked and PASSED using:
```
for vault_id in default rtk ec2 all
do
        echo ==== $vault_id ======
        via_op=$(bin/ansible-vault-client --vault-id $vault_id)
        echo "VIA_OP  :$via_op"
        if [ -e .keybase ] ; then
            file=".${vault_id}-vault-pass"
            [ -s "$file" ] || file=.vault_pass.txt
            echo "VIA_FILE:$(cat "$file")"
        fi
done
```

* Merge and tell the team: 'Run make clean && make requirements and confirm everything works. If your op setup is good, make reads from 1Password silently. If not, Keybase still serves you.'

Fails if not signed in to 1Password:

```
$ make clean && make requirements
rm -rf .venv roles/external site.retry collections .keybase .make
rm -rf terraform/.terraform
rm -f terraform/secrets.auto.tfvars
WARN: OAF 1Password not reachable (account=oaforgau). Will fall back to Keybase symlinks where possible.
bin/fetch-terraform-pem
fetch-terraform-pem: failed to read terraform.pem from 1Password
  Sign into the OAF account: op signin --account oaforgau
make: *** [Makefile:110: terraform.pem] Error 1
(Works if signed in)
```

benrfairless:
@ianheggie-oaf @jamezpolley I've pushed changes addressing the "remove the requirement that Keybase keeps working" point — Keybase is now fully removed and 1Password is the sole source. Summary of what changed since the last review:

**Keybase runtime fallback removed**
- `bin/ansible-vault-client` no longer falls back to the `.*-vault-pass` symlinks. If `op` can't read the passphrase it now fails cleanly with a "sign into 1Password" message and a non-zero exit. (Verified: signed in → prints the passphrase; signed out → fails, even with the symlinks still on disk.)
- `bin/fetch-terraform-pem` drops the Keybase framing and just requires the 1Password CLI.

**`make` fails fast instead of cryptically**
- `op-check` is now **fatal** rather than a warning. This fixes the `make clean && make requirements` behaviour you flagged — instead of warning and then dying later at `terraform.pem`, it now stops immediately at `op-check` with a clear "sign in: `op signin --account oaforgau`" message.

**Legacy Keybase machinery deleted**
- Removed the `keybase` / `.keybase` Makefile targets, `bin/headless-keybase.sh`, and the four committed `.*-vault-pass` symlinks.
- `packer/planningalerts.pkr.hcl` previously passed those symlinks as `--vault-id`; it now points at the `bin/ansible-vault-client` dispatcher, matching `ansible.cfg`. ⚠️ Note: the Packer build host therefore needs to be signed into the OAF 1Password account now.
- Updated `README.md`, `ansible.cfg`, `site.yml`, and `.gitignore` to drop the fallback story and the now-obsolete `ANSIBLE_VAULT_IDENTITY_LIST` override that referenced the symlink files.

**PR description updated** to match, and I corrected a stale claim in it: the Cloudflare and Linode API tokens are shared service tokens rendered into `secrets.auto.tfvars` from 1Password by `make tf-secrets` (alongside `rds_admin_password`) — they are not sourced from per-operator `CLOUDFLARE_API_TOKEN` / `LINODE_TOKEN` env vars.

Net effect for operators: you must be in the OAF `DevOps` vault (and `RTK Devops` for RTK) and signed into `op` before running `make`. There is no Keybase path any more.

bin/ansible-vault-client
```diff
@@ -0,0 +1,42 @@
...
+    *) shift ;;
+  esac
+done
+
+case "$vault_id" in
+  default) op_ref="op://nal3r3nlyo2sdqzegvdvlmvp5e/m2ccf6rvruxlyhevcovwher4cy/password" ;;
+  all)     op_ref="op://nal3r3nlyo2sdqzegvdvlmvp5e/mpwsqfdmg7xinu46zmiqesn2yu/password" ;;
+  ec2)     op_ref="op://nal3r3nlyo2sdqzegvdvlmvp5e/emvvhq7denilwx2nesg45fkwrm/password" ;;
+  rtk)     op_ref="op://nal3r3nlyo2sdqzegvdvlmvp5e/33howvcuao7gomvqpivs7dtzoy/password" ;;
+  *) echo "ansible-vault-client: unknown vault id '$vault_id'" >&2; exit 1 ;;
```
jamezpolley, line 33: This line of code will never be executed. If any unknown vault id is passed in, line 15 sets vault_id to "default" and nothing in 16-22 ever changes it - so line 25 will run.

This means that things like "ansible-vault-client --help" return a secret instead of the expected help. It means that passing in an unexpected string will be hard to detect, because this script will not return an error.

bin/ansible-vault-client
```diff
@@ -0,0 +1,42 @@
...
+export OP_ACCOUNT
+
+vault_id="default"
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --vault-id) vault_id="$2"; shift 2 ;;
+    --vault-id=*) vault_id="${1#*=}"; shift ;;
+    *) shift ;;
+  esac
+done
```
jamezpolley:
```suggestion
if [ $# -eq 0 ]; then
  vault_id="default"
else
  while [ $# -gt 0 ]; do
    case "$1" in
      --vault-id) vault_id="$2"; shift 2 ;;
      --vault-id=*) vault_id="${1#*=}"; shift ;;
      *) vault_id+=$1; break ;;
    esac
  done
fi
```

bin/ansible-vault-client
jamezpolley: Suggest change above:
- preserves the "no vault id -> return default" behavior, which I assume is desired
- Aborts with an explicit error on the first unexpected piece of input (ie, anything other than --vault-id followed by the value)

If possible, I'd prefer the "no input" case to also return an error, in order to avoid printing out an unexpected secret, but I assume this would require changes else. This is good enough to be going forward.

jamezpolley (changes requested):
I haven't tested this extensively, just given it the quick test suggested by the docs - I ran "make check-servername" on a random server, everyting worked fine.

I did get some unexpected results from running ansible-vault-client by hand which turned out to be caused by the script treating bad input as though it was a valid request for the default secret.

README.md
```diff
@@ -187,33 +190,46 @@ If it makes sense we might move cuttlefish and morph.io to AWS as well.
...
-  - Alternately symlink from your own copy of the keybase files
-    - If Keybase isn't working for you, any technique you have to put the right value into the right file will be fine,
-- You may override `vault_identity_list` in [ansible.cfg](https://github.com/openaustralia/infrastructure/blob/master/ansible.cfg) to point at your new location,
-  or list just the files you have access to.
-- Consider using the direnv command and setting
-  `export ANSIBLE_VAULT_IDENTITY_LIST=".vault_pass.txt,ec2@.ec2-vault-pass,all@.all-vault-pass"`
-  in your `.envrc` file if you don't have all the perms required.
+- Ansible reads each vault passphrase from the OAF 1Password account via [bin/ansible-vault-client](bin/ansible-vault-client). See [Add the Ansible Vault password](#add-the-ansible-vault-password) below.
 
-**Capsitrano (in project repos)**
```
jamezpolley, line 216: 🤣🤣🤣🤣🤣

bin/rotate-vault-passphrase
```diff
@@ -0,0 +1,196 @@
...
+fi
+
+# 3. Update 1Password FIRST (before touching any files) using a JSON
+#    template piped via stdin. The new passphrase never appears on a
+#    command line, so it can't be observed via `ps` / /proc/<pid>/cmdline.
+echo "  updating 1Password item ($op_item)..."
+op item get "$item_uuid" --account "$OP_ACCOUNT" --format=json \
+  | jq --rawfile pw "$new_pw" \
+       '.fields |= map(if .id == "password" then .value = ($pw | rtrimstr("\n")) else . end)' \
+  | op item edit "$item_uuid" --account "$OP_ACCOUNT" >/dev/null
```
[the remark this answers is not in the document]
benrfairless: This is a false positive — `op item edit` does support piped JSON on stdin per its own `--help`: "You can also edit an item using piped input: `cat updatedLogin.json | op item edit oldLogin`". Verified against `op` 2.36.0-beta.03. No change needed here; the code as written is correct.

benrfairless:
Pushed three commits addressing the outstanding review feedback:

**`68130dc7` (was `2d2f7d05`, force-pushed nothing, just amended locally before this was on origin)**
- `bin/ansible-vault-client`: fixes @jamezpolley's bug report. No-args now prints a usage error and exits non-zero instead of silently returning the `default` passphrase, so `ansible-vault-client --help` or any unexpected input fails loudly rather than leaking a secret.
- `bin/rekey-yaml`: fixes the Sentry MEDIUM finding. A malformed `!vault \|` block (no body, or a body that isn't indented) now logs a warning and skips just that block with `next`, instead of `break`-ing out of the whole file and silently truncating everything after it.
- `bin/rotate-vault-passphrase`: the `rtk` entry was still pointing at the old `RTK Devops` vault UUID (`vya2dzj3…`) rather than the shared `DevOps` vault (`nal3r3nlyo2sdqzegvdvlmvp5e`) that `bin/ansible-vault-client` and everything else actually reads from. Fixed to match.

**`93e14c99` - Fix packer vault-id label for default passphrase**
Knock-on effect of the `ansible-vault-client` fix above: `packer/planningalerts.pkr.hcl` was passing a labelless `--vault-id ../bin/ansible-vault-client`, which invokes the script with zero arguments. That used to work by accident (it hit the old no-args-returns-default bug), but now hits the new hard error, which would have broken the Packer build. Tagged it `default@../bin/ansible-vault-client` to match the other two vault-id args and `ansible.cfg`. Verified the args-passing behaviour directly against `ansible-vault` 2.10 (labelless vs labelled) to confirm this was a real break, not a hypothetical one.

**`1d3cc779` - docs: correct README claim about RTK Devops vault**
The README told new operators to get added to "DevOps (and RTK Devops for Right To Know admins)". That's stale: the `rtk` passphrase currently lives in the same `DevOps` vault as the other three, per `6f665fec`'s "for now" (the `rotate-vault-passphrase` fix above confirms this is also what the code expects). Updated the two affected sections rather than moving the actual 1Password item, since that's an access-control decision, not a docs bug, happy to move the secret into its own vault instead if that's the preferred fix, in which case the vault/item UUIDs in `ansible-vault-client` and `rotate-vault-passphrase` would need updating to match.

Also replied inline to Sentry's CRITICAL finding on `op item edit`, it's a false positive; piped stdin JSON is explicitly supported per `op item edit --help`.

Re-requested review from @ianheggie-oaf and @jamezpolley.

jamezpolley (approved):

PR #534, opened 2026-05-31
Feature/tag git when servers are updated

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/oaf-internal/issues/243

## What does this do?

tags git repo when ansible or terraform provisioning is performed via make

## Why was this needed?

Ian's KPIs on release cadence

### DIFF
--- a/bin/tag-provisioning
+++ b/bin/tag-provisioning
@@ -10,6 +10,7 @@ fi
 
 if [ $# -ne 4 ]; then
   echo "Usage: bin/tag-provisioning [--wip] <target> <stage> <tags> <skip_tags>"
+  echo "       Should be called with --wip first, then do action then call without --wip to replace wip tag"
   echo "       Use empty strings for unused args, eg: bin/tag-provisioning oaf \"\" \"\" \"\""
   echo "       Intended to be called from make apply-* and tf-apply targets"
   exit 1
@@ -40,12 +41,12 @@ build_tag() {
 if $WIP; then
   TIMESTAMP=$(date -u +%Y%m%d%H%M%S)
   if [ -f "$WIP_FILE" ]; then
-    echo "Discarding previous wip timestamp: $(cat "$WIP_FILE")"
+    echo "Discarding previous wip timestamp (presumably provisioning failed): $(cat "$WIP_FILE")"
   fi
   echo "$TIMESTAMP" > "$WIP_FILE"
 else
   if [ ! -f "$WIP_FILE" ]; then
-    echo "Error: Missing $WIP_FILE with timestamp!"
+    echo "Error: Missing $WIP_FILE with timestamp! Call with --wip before the actual provisioning command so attempts are tagged!"
     exit 1
   fi
   TIMESTAMP=$(cat "$WIP_FILE")
--- a/bin/tag-provisioning
+++ b/bin/tag-provisioning
@@ -66,7 +66,7 @@ fi
 if ! $WIP; then
   rm -f "$WIP_FILE"
   echo "Removing: $WIP_TAG"
-  git tag -d "$WIP_TAG"
+  git tag -d "$WIP_TAG" 2>/dev/null || true
   if $HAS_REMOTE; then
     git push origin --delete "$WIP_TAG"
   fi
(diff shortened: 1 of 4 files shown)


bin/tag-provisioning
```diff
@@ -0,0 +1,74 @@
...
+
+TARGET="$1"
+STAGE="$2"
+TAGS="$3"
+SKIP_TAGS="$4"
+
+mkdir -p tmp
+WIP_FILE="tmp/wip-tag-timestamp"
+
+HAS_REMOTE=$(git remote | grep -q origin && echo true || echo false)
```
Br3nda: -x means "exact match" so it won't trigger when you have a remote name "old-origin" or "bad-origin" etc.
(just in case)

```suggestion
HAS_REMOTE=$(git remote | grep -xq origin && echo true || echo false)
```

Br3nda: or maybe this is better:
```suggestion
if git remote | grep -qx origin; then HAS_REMOTE=true; else HAS_REMOTE=false; fi
```

bin/tag-provisioning
```diff
@@ -0,0 +1,74 @@
...
+    echo "Discarding previous wip timestamp: $(cat "$WIP_FILE")"
+  fi
+  echo "$TIMESTAMP" > "$WIP_FILE"
+else
+  if [ ! -f "$WIP_FILE" ]; then
+    echo "Error: Missing $WIP_FILE with timestamp!"
+    exit 1
+  fi
+  TIMESTAMP=$(cat "$WIP_FILE")
+fi
```
Br3nda, line 55: maybe some comments here to exaplin the logic around WIP

bin/tag-provisioning
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+set -euo pipefail
+trap 'echo "ERROR: tag-provisioning failed at line $LINENO (exit $?)" >&2' ERR
```
Br3nda, line 3: maybe check that this is being run from the root of the repo, because i think it might fail if you're not.

bin/tag-provisioning
```diff
@@ -0,0 +1,74 @@
...
+echo "Tagging: $TAG"
+git tag "$TAG"
+if $HAS_REMOTE; then
+  git push origin "$TAG"
+fi
+
+if ! $WIP; then
+  rm -f "$WIP_FILE"
+  echo "Removing: $WIP_TAG"
+  git tag -d "$WIP_TAG"
```
Br3nda: if the tag doesn't exist then still succeed
```suggestion
  git tag -d "$WIP_TAG" 2>/dev/null || true
```

bin/tag-provisioning
```diff
@@ -0,0 +1,74 @@
...
+  if $HAS_REMOTE; then
+    git push origin --delete "$WIP_TAG"
```
Br3nda: same here
```suggestion
    git push origin --delete "$WIP_TAG" 2>/dev/null || true
```

bin/tag-provisioning
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+set -euo pipefail
+trap 'echo "ERROR: tag-provisioning failed at line $LINENO (exit $?)" >&2' ERR
+
```
Br3nda, line 4: something like:
```suggestion
# Anchor to repo root so WIP file is consistent regardless of CWD
REPO_ROOT=$(git rev-parse --show-toplevel)
cd "$REPO_ROOT"
```

Br3nda (commented):
added suggestions

benrfairless:
Taking myself off as a reviewer, happy to review when the comments have been addressed.

bin/tag-provisioning
ianheggie-oaf: Elaborated in the messages to hopefully make clearer

bin/tag-provisioning
ianheggie-oaf: Have enforced it, though its unlikely to be needed, since Makefile assumes CWD is the project root and nearly every command will fail if it isn't

bin/tag-provisioning
ianheggie-oaf: Used: `cd "$(dirname "$0")/.." || exit 1`

ianheggie-oaf:
@Br3nda and @benrfairless - I have updated based on Brenda's feedback - ready for 2nd review

benrfairless (approved):

PR #538, opened 2026-06-06
Adding composer for TWFY php

Br3nda:
## Relevant issue(s)
https://github.com/openaustralia/openaustralia/issues/863

## What does this do?

Adds composer to our server that run TWFY php

## Why was this needed?

to install php dependencies. We've moved to using composer to manage php dependencies.


benrfairless (approved):
LGTM 🚀

PR #539, opened 2026-06-10
fix xapian via ansible, in openaustralia.org.au

Br3nda:
## Relevant issue(s)

closes
 * https://github.com/openaustralia/openaustralia/issues/864

needs
 * https://github.com/openaustralia/infrastructure/pull/540

## What does this do?
Adds Xapian PHP bindings installation in the `openaustralia` role:

- Pins `libxapian-dev` and `xapian-tools` (from `ppa:xapian/backports`) to a specific upstream version declared in `group_vars/openaustralia.yml` (`xapian_version`, currently `1.4.22`), instead of taking whatever the PPA happens to ship.
- Verifies the downloaded `xapian-bindings-<version>.tar.xz` against a pinned SHA256 (`xapian_bindings_sha256`), retiring the existing `FIXME: handle gpg checksum` comment.
- Makes the build-from-source chain idempotent via a version-keyed stamp file at `/var/lib/xapian-bindings-installed-<version>`. Re-running the role no longer redownloads, reconfigures or rebuilds when the pinned version is already installed; bumping the pin automatically triggers a rebuild.
- Add `extension=xapian.so` / `extension=mysqli` lines in `templates/php.ini`
- fixes ansible lint failing on a yml file that isn’t actually for ansible

### DIFF
--- a/roles/internal/openaustralia/tasks/main.yml
+++ b/roles/internal/openaustralia/tasks/main.yml
@@ -193,6 +193,7 @@
       - libmagickcore-dev
       - libmagickwand-dev
       - libmysqlclient-dev
+      - libsearch-xapian-perl
       - libxml-rss-perl
       - libxml-simple-perl
       - libxml-twig-perl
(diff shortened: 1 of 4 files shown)


roles/internal/openaustralia/tasks/main.yml
```diff
@@ -227,29 +249,39 @@
...
 - name: Download xapian bindings source
   tags: xapian
+  when: not xapian_stamp.stat.exists
   get_url:
     url: https://oligarchy.co.uk/xapian/{{ xapian_version }}/xapian-bindings-{{ xapian_version }}.tar.xz
     dest: /home/deploy
     owner: deploy
     group: deploy
-    # FIXME: handle gpg checksum: checksum: sha256:b15ca7984980a1d2aedd3378648ef5f3889cb39a047bac1522a8e5d04f0a8557
+    checksum: "sha256:{{ xapian_bindings_sha256 }}"
```
Br3nda, line 280: i still didn't implement GPG here, but i wanted to make ansible safe to apply

ianheggie-oaf (approved):
Approved - LGTM, updated description to include a fix for ansible lint

Br3nda:
I've got a better idea for fixing the yaml-lint over in #544

Issue #523, opened 2026-05-25
Running ansible on Right to Know removes rbenv


benrfairless:
**Describe the bug**
rbenv is a requirement for Right to Know, however running `make check-righttoknow STAGE=staging` tries to nuke rbenv.

**To Reproduce**
A. Run 

       make check-righttoknow STAGE=staging

B. Wait for the tasks to get to the rbenv removal stage

**Expected behavior**
Right to Know uses rbenv and therefore it should not be uninstalled when running ansible

**Screenshots/Output**
N/A

**Additional context**
This appears to have been introduced in 54cdf2d76f434b709def081324b18adf21c15c7e

Related:
* https://github.com/openaustralia/infrastructure/issues/477

benrfairless:
Ruby configuration for Right to Know is done here:
https://github.com/openaustralia/infrastructure/blob/8a4989a90786b9f899d28af45bf94b5d730f45bd/roles/internal/righttoknow/meta/main.yml#L124-L133

When I check [openaustralia](https://github.com/openaustralia/infrastructure/blob/main/roles/internal/openaustralia/meta/main.yml) I can't find a similar entry.

benrfairless:
@ianheggie-oaf it appears that we are getting into a bit of a mess here. I note that we currently run a removal job every time we run the main playbook:
https://github.com/openaustralia/infrastructure/blob/8a4989a90786b9f899d28af45bf94b5d730f45bd/site.yml#L171-L199

We need to revisit this with a goal to reduce deployment time while reducing the need to remember where we need to configure things in various roles, especially if it's going to uninstall somethhing.

benrfairless:
Reduced priority on this as we have implemented a workaround in #542

PR #542, opened 2026-06-10
Exclude Right to Know from Ruby Manager Tasks

benrfairless:
## Relevant issue(s)
#523

## What does this do?
Excludes the Right to Know (RTK) application from the Ruby manager tasks in the Ansible playbook.

## Why was this needed?
Running the `make check-righttoknow STAGE=staging` command was inadvertently removing rbenv, which is a requirement for RTK. This change prevents the removal of rbenv when executing the playbook.

1. Test by running `make check-righttoknow STAGE=all TAGS=ruby`
2. Ensure that there are no referenes to the "Setup ruby managers and install versions" or "Remove unused ruby managers" tasks. 

This is a temporary workaround to address the issue until a more permanent solution can be implemented.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -168,10 +168,10 @@
     - role: geerlingguy.swap
       when: swap_file_size_mb is defined or (swap_file_state | default('present')) == 'absent'
 
-- hosts: all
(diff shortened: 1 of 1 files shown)


ianheggie-oaf (approved):
LGTM - quick workaround

PR #553, opened 2026-07-03
DNS Domain Verification - Apple, GitLab, Mintlify

benrfairless:
## Description
This change performs various DNS Domain verifications for Apple, GitLab and Mintlify.

## Motivation and Context
This change is required to verify that we own our domain.

## How Has This Been Tested?
Testing was performed using `make tf-plan` which shows the changes are what is expected.

## Types of Changes

## Checklist:


ianheggie-oaf (dismissed):
I agree with sentry that the acme challenge value is wrong.

I was also confused by the description - to me rather than:

This change is required to allow users to verify their domains through DNS records, improving the overall user experience and ensuring domain ownership verification

Instead it

1. Points handbook to mintlify
2. Validates we own the domain name so mintlify can issue an SSL certificate

You should copy the value from the mintlify dashboard into value (possibly via a variable, though the value shouldn't change it is sourced externally and worth commenting where it came from)

benrfairless (commented):
- Updated PR with detail on exactly what I'm doing
- Update ACME challenge (good pick up!)
- Re-requested review

jamezpolley (approved):

PR #600, opened 2026-07-23
Fix theyvoteforyou_foreman sudoers task using template module

### LINKED ISSUE #598: theyvoteforyou_foreman sudoers task fails: uses `copy` but file moved to `templates/`
**Describe the bug**

The Ansible task **"Allow deploy user to control application services"** in the `theyvoteforyou` role fails on every run (both `--check` and a real apply).

The task references `theyvoteforyou_foreman` via the `copy` module, which only searches the role's `files/` directory. Commit 25d76f8b ("Refactor ruby version management to be generic and move to site.yml") relocated that file from `files/` to the role's `templates/` directory, so `copy` can no longer find it.

Because the task fails, the play halts and **every subsequent `theyvoteforyou` task is skipped**, so the role can no longer complete a run against the server.

Task location: `roles/internal/theyvoteforyou/tasks/main.yml` (lines 267-273).

**To Reproduce**

1. Given I have the infrastructure repo checked out with 1Password/AWS access
2. When I run `make check-theyvoteforyou ANSIBLE_TAGS=theyvoteforyou`
3. Then I expected the play to run to completion
4. But instead the play fails at the "Allow deploy user to control application services" task with the error below

**Expected behavior**

The task locates `theyvoteforyou_foreman` and completes, allowing the rest of the role to run.

**Screenshots/Output**

```
TASK [theyvoteforyou : Allow deploy user to control application services] ******
fatal: [srv.theyvoteforyou.org.au]: FAILED! => {"changed": false, "msg": "Could not find or access 'theyvoteforyou_foreman'\nSearched in:\n\t.../roles/internal/theyvoteforyou/files/theyvoteforyou_foreman\n\t... on the Ansible Controller.\nIf you are using a module and expect the file to exist on the remote, see the remote_src option"}
```

**Additional context**

- The file was moved `files/` → `templates/` in commit 25d76f8b.
- It now lives at `roles/internal/theyvoteforyou/templates/theyvoteforyou_foreman`.
- Likely fix: switch the task from the `copy` module to the `template` module (the file lives in `templates/` now, even though it currently contains no Jinja variables).

---
_Filed with AI assistance (Claude Code, model claude-opus-4-8). Findings verified against the repo and a live `--check --diff` run; no changes were applied._

benrfairless:
## Description

The Ansible task **"Allow deploy user to control application services"** in the `theyvoteforyou` role used the `copy` module, which only searches the role's `files/` directory. Commit `25d76f8b` moved `theyvoteforyou_foreman` into the role's `templates/` directory, so `copy` could no longer find it. The task failed on every run, halting the play and skipping all subsequent `theyvoteforyou` tasks.

This switches the task from `copy` to the `template` module (which searches `templates/`), matching the equivalent task in the `planningalerts` role. The file now also contains a `{{ bundle_path }}` variable that needs rendering, so `template` is required rather than optional. The `dest` is set to an explicit filename because `template`, unlike `copy`, does not append the source basename when `dest` is a directory.

## Motivation and Context

Fixes #598. Without this, the `theyvoteforyou` role cannot complete a run against the server.

## How Has This Been Tested?

  - Ran `ansible-lint roles/internal/theyvoteforyou/tasks/main.yml` — passes (only the expected informational notice about the custom yamllint config).
  - Confirmed `bundle_path` is defined in `group_vars/all.yml`, so the template renders.
  - Verified the fix mirrors the working `planningalerts` `deploy_service_control` task (`template` + explicit `dest` filename).

N/A

## Types of Changes

- [ ] Documentation

## Checklist:

---

_Assisted-by: Claude Code/claude-opus-4-8. AI-generated changes reviewed by the PR author_


ianheggie-oaf (changes requested):
Thanks for fixing my mistake, but there is (now) a merge conflict to be resolved.

ianheggie-oaf (changes requested):
theyvoteforyou_foreman should have a .j2 extension to follow recently established convention for template files

benrfairless:
@ianheggie-oaf I think this fixes #668 as well - can you check and confirm?

ianheggie-oaf (approved):
Approved - LGTM

PR #606, opened 2026-07-24
Post upgrade/clean

Br3nda:
## Description

After upgrading to Mysql 8.4 we want to turn off allowing major version upgrades

I also found some mysql 5 config we can remove.

## Motivation and Context

motivation is to be clean and tidy.


terraform/database.tf
```diff
@@ -117,7 +65,7 @@ resource "aws_db_instance" "maindb" {
   # TODO: Remove this (or set back to false) once the upgrade has completed,
   # so an accidental engine_version bump in future can't trigger a surprise
   # major version upgrade.
```
Br3nda:
```suggestion
```

ianheggie-oaf (approved):
LGTM - I assume terraform will give an informative message when it rejects the next major upgrade why it did so.

PR #612, opened 2026-07-28
Add local Active Storage service to RightToKnow storage.yml

benrfairless:
## Description

Adds a `local:` Disk service to the RightToKnow `storage.yml` that ansible copies to `/srv/www/<stage>/shared/storage.yml`.

```yaml
local:
  service: Disk
  root: <%= Rails.root.join('storage/local') %>
```

The root matches Alaveteli's own `config/storage.yml-example`. Nothing is expected to be written there (see below), but the key has to exist for the app to boot.

## Motivation and Context

`cap staging deploy` of Alaveteli 0.45.8.0 fails at `deploy:assets:precompile`:

```
KeyError: Missing configuration for the local Active Storage service.
Configurations available for the raw_emails_disk, raw_emails_production, ...
attachments_test, and attachments services.
  activestorage-8.0.1/lib/active_storage/service/registry.rb:18:in `block in fetch'
  activestorage-8.0.1/lib/active_storage/engine.rb:153
  app/models/foi_attachment.rb:42:in `<class:FoiAttachment>'
```

Alaveteli's `config/environments/production.rb` sets `config.active_storage.service = :local`, added upstream by the Rails 8 `rails app:update` (alaveteli commit `8b49ca12a`). Active Storage resolves that key **eagerly** — as soon as `ActiveStorage::Blob` loads, which `has_one_attached` in `app/models/foi_attachment.rb` triggers during boot. So it fails even though nothing actually stores blobs in `:local`.

`config/storage.yml` is gitignored in alaveteli and supplied via `SHARED_FILES`, symlinked from `/srv/www/<stage>/shared/storage.yml`, which this role copies. Our copy predates the Rails 8 change and defined only the named `raw_emails` / `attachments` services, so 0.45.8.0 cannot boot on either stage.

**Production is affected too.** It hasn't broken yet only because `current` is an April release (`7b8014c45`) whose `production.rb` has no `active_storage` lines at all. The next `cap production deploy` at 0.45.8.0 fails identically. This file feeds both stages via `with_items`, so this one change covers both — production's shared copy updates when the playbook next runs against `righttoknow_production`.

### Where the files land, and what goes in there

`release/storage` → `/srv/www/<stage>/shared/storage` → `/data/<stage>/storage`, so `storage/local` resolves to `/data/<stage>/storage/local`: the persistent volume, beside the existing `attachments/` and `raw_emails/`, and inside the restic backup set (`source: /data`, excluding only `**/bundle` and `**/cache`). No backup gap.

In practice nothing is written there today. A blob census on staging returns `raw_emails: 9`, `attachments: 15` and zero `local` — both attachment types pin an explicit service (`has_one_attached :file, service: :attachments`). The only thing that would use the default is Action Text, whose `ActionText::RichText` declares `has_many_attached :embeds` with no `service:` override; that is currently impossible because `action_text_rich_texts` does not exist in the database.

## How Has This Been Tested?

Applied to staging with `ansible righttoknow_staging -m copy --become --diff`, whose diff confirmed exactly the one added block and no other drift. A full `--check --diff --tags alaveteli` run is not usable here — it fails early on `AnsibleUndefinedVariable: rds_database is undefined`, because the RDS fact-gathering play is skipped by the tag filter.

Verified in the release that had already failed, since its `config/storage.yml` symlinks to the shared file:

| Command (in `releases/20260728121403`) | Before | After |
| --- | --- | --- |
| `bundle exec rake environment RAILS_ENV=production` | `KeyError` | exit 0 |
| `bundle exec rake assets:precompile` | `KeyError` | exit 0, theme assets included |

No automated tests cover this file. CI will run on this PR.

Unrelated and pre-existing: the deploy logs a Flipper warning (`rails generate flipper:update` / `rails db:migrate`) on stderr. It appears both before and after this change and does not affect the exit status.

Not applicable — server configuration change.

## Types of Changes

- [ ] Documentation

## Checklist:

Documentation change not required: the added block is commented inline explaining why the key must exist, and the file otherwise tracks upstream's `storage.yml-example`.

---

**AI use:** I diagnosed this failure, wrote the change, and applied and verified it on staging with the help of Claude Code (model `claude-opus-5`). I have reviewed the change and the reasoning above, and I'm responsible for it.

### DIFF
--- a/roles/internal/righttoknow/files/storage.yml
+++ b/roles/internal/righttoknow/files/storage.yml
@@ -20,6 +20,14 @@
 #   project: ''
 #   bucket: ''
 
+# Alaveteli's models attach to the named :raw_emails and :attachments services
+# below, but config.active_storage.service in config/environments/production.rb
+# is :local, and Active Storage resolves that key eagerly when
+# ActiveStorage::Blob loads. It must exist or the app fails to boot.
+local:
+  service: Disk
+  root: <%= Rails.root.join('storage/local') %>
+
 ## Raw Emails ##
 
 raw_emails_disk: &raw_emails_disk


ianheggie-oaf (approved):
LGTM - Given its been applied and fixes the issue, its approved

PR #624, opened 2026-08-07
Infrastructure upgrades required for Alaveteli 0.46.7.0

benrfairless:
## Description

Infrastructure changes required for Alaveteli 0.46.7.0 (openaustralia/alaveteli#55). Merged up to current `main`, which has absorbed part of what this branch originally carried, so the remaining delta is two files:

- `bundler_version` 2.6.2 → 2.7.2 in `roles/internal/righttoknow/defaults/main.yml`, matching `BUNDLED WITH` in the 0.46.7.0 `Gemfile.lock`. An older bundler in the same major series does still install the bundle, so this is not strictly load-bearing, but matching it avoids the version-mismatch warning and keeps a rebuilt server consistent with the one we have. Apply before the first Capistrano deploy of the new code. The surrounding comment is reworded so the `BUNDLED WITH` reference points at 0.46.7.0's 2.7.2 lockfile rather than 0.45.8.0's 2.6.2 one.
- `roles/internal/righttoknow/tasks/cron.yml` adds the two cron jobs from the regenerated 0.46 crontab, both wrapped in `sentry monitor run` the way #1004 wrapped every other job in this file: `users-purge-limited` at 02:50 and `users-destroy-unused` at 02:53, each passing its own schedule to `-s`, `--check-in-margin 60 --max-runtime 120 --timezone Australia/Sydney` to match the other daily jobs, and `rtk-cron-lock ../<name>.lock` for the lockfile convention the other `bin/rails` jobs use. Both slots are free: the neighbouring jobs sit at 02:38, 02:40, 02:43, 02:48 and 02:54.

Both jobs are scoped to the `righttoknow_staging` group only for now, so a production provision deploys neither; the production rollout is a decision for the 0.46.7.0 cutover PR. On staging, `users-purge-limited` is active. `users-destroy-unused` ships `disabled: true`: `last_sign_in_at` is nil for every account predating the upgrade, so its first run would sweep every "unused" account older than two years. It gets enabled in a follow-up once the dry-run count on staging is reviewed and signed off on openaustralia/righttoknow#1031. A disabled entry is commented out in the crontab, so it never runs and never checks in, which means no Sentry monitor is registered for it until the entry is enabled - enabling it later brings the monitor with it.

No `general.yml.j2` change. The three settings 0.46.0.0 introduces - `READ_ONLY_FEATURES`, `ENABLE_PUBLIC_ANNOTATIONS`, `ENABLE_USER_TO_USER_MESSAGING` - landed on `main` already in 5088b0c5 ("Sync general.yml keys with upstream examples"), so the commit that added them here has been dropped as redundant. The template's key set was re-checked against `config/general.yml-example` at the 0.46.7.0 tag: every upstream key is present, with the only extras being our three `SENTRY_*` keys. Two deliberate divergences from upstream remain, both pre-existing on `main`: `BLOCK_SPAM_USER_MESSAGES: true` where upstream defaults to `false` (set in 5088b0c5 to match our other `BLOCK_SPAM_*` toggles), and the `READ_ONLY` comment block, which still describes the default as `nil` where 0.46.7.0's example says `''`. The latter is comment text only - the value we set, `READ_ONLY: ''`, matches upstream - so it has been left alone.

No Ruby change: righttoknow#1045 moved staging and production to 3.4.10 on `main`, so this PR no longer touches `ruby_version_staging`/`ruby_version_production`. No package changes, and `sidekiq.yml` is unchanged between the tags.

## Motivation and Context

Final stage of the staged upgrade tracked in openaustralia/righttoknow#1005; this jump is openaustralia/righttoknow#1031, which carries the staging deploy runbook.

## How Has This Been Tested?

Static checks only so far. `yamllint` on the two changed files reports no new issues - the four extra line-length warnings on `cron.yml` are the two new `job:` lines and their `with_items:` lines, which are the same shape as every other entry in the file. GitHub Actions is green across the board, including `ansible-lint` (which was not run locally). Nothing has been run against a real host yet: `make check-righttoknow STAGE=staging` before applying.

## Types of Changes

## Checklist:

---
Assisted-by: Claude Code:claude-opus-5

The merge up to `main`, the conflict resolutions and the cron rewrite were drafted with AI assistance and human-reviewed before submission.


ianheggie-oaf (approved):
Approved on the basis its going to be checked in staging before it goes to production - some friction that an urgent fix might get mixed in with this to be tested in staging, but it is manageable

PR #625, opened 2026-08-09
Add CLAUDE.md, docs/DECISIONS.md, and a suggested "assembling" term

ianheggie-oaf:
## Description

Adds `CLAUDE.md`: 
* setup, 
* common `make`/terraform/ansible commands,
* architecture (terraform/ansible split,   vault ID scheme, `.j2` template convention), and 
* AI-tool working conventions (production-change caution, secrets handling, git staging via `GITGUI_MSG`, PR/issue/AI-disclosure conventions), 
* adapted from the theyvoteforyou repo's `CLAUDE.md`.

Adds `docs/DECISIONS.md`:
-  a cross-cutting decisions log (same convention as theyvoteforyou's),
- seeded with an  entry for the planned move of local Ansible testing from Vagrant to Docker/Compose.

Adds a third term, "assembling", to `README.md` and `CLAUDE.md` existing terminology notes, for using Terraform  to create/update infrastructure before Ansible provisions the server. Marked explicitly as a suggestion, not yet  in wide use, alongside the existing "provisioning"/"deployment" terms.

## Motivation and Context

No CLAUDE.md existed for this repo, so AI tools had no repo-specific guidance (commands, architecture, or the
"never run apply-*/tf-apply* without explicit go-ahead" caution that matters most here). Docs-only change, no
behaviour affected.

## How Has This Been Tested?

- [x] Checked affected area manually on my own / staging system (read through the rendered Markdown; verified
      referenced commands/paths/targets against the current Makefile, README, and repo layout)
- [ ] Ran automated tests on my own system

## Types of Changes

- [x] Documentation

## Checklist

---

Parts of this contribution (drafting CLAUDE.md, docs/DECISIONS.md, and the terminology wording) were generated
or assisted by Claude Code (claude-sonnet-5), reviewed by @ianheggie-oaf before submission.


CLAUDE.md
benrfairless, line 1: Can we please re-name this to AGENTS.md, and create a CLAUDE.md and a copilot-instructions.md to point to the new AGENTS.md file.

benrfairless (changes requested):

CLAUDE.md
```diff
@@ -0,0 +1,164 @@
...
+
+- If something here doesn't match what you're consistently seeing in the repo, flag the mismatch and ask which
+  needs fixing (so it's fixed once and for all), presenting fixing the code/config as the easy default choice and
+  updating this file as the alternative.
+- When a commit message body covers more than one distinct point, use a markdown bullet list rather than one
+  flowing paragraph; it's easier to scan and review.
+- Check `docs/DECISIONS.md` for past cross-cutting decisions before assuming in an unfamiliar area of the repo; add
+  a new entry there (rather than repeating it in multiple places) when a decision spans multiple files/roles/modules
+  instead of belonging as a comment in one.
+- `apply-*`, `tf-apply*`, `tf-apply-target`, `letsencrypt`, and `bin/rotate-vault-passphrase` change real production
```
benrfairless: This is the one I'd call blocking. The list is missing the three most dangerous targets in the repo:

- `make all` (`Makefile:174`) runs `site.yml` against every host, no `--check`
- `make retry` (`Makefile:180`)
- `make update-github-ssh-keys` (`Makefile:296`) runs `site.yml --tags userkeys` against all hosts, so it governs who can SSH in

`letsencrypt` is on the list and `make all` isn't, even though both run `ansible-playbook` with no `--check` and no `-i`, so it reads as an oversight rather than a deliberate call. Line 55 also documents `make all` neutrally, which together tells an agent it's an ordinary command.

Worth adding a clause too: unlike `apply-*`, none of these three is wrapped by `bin/tag-provisioning`, so the lingering `wip-*` signal described at line 74 doesn't apply to them.

CLAUDE.md
```diff
@@ -0,0 +1,164 @@
...
+make help                              # full list of targets with descriptions
+make lint                              # yaml-lint + template-check + ansible-lint + tf-check-fmt + tf-validate
+make yaml-lint                         # yamllint on roles/ and site.yml only
+make ansible-lint                      # ansible-lint on roles/ and site.yml only
+make template-check                    # fails if any role's templates/ file doesn't end in .j2
+
+# Ansible provisioning (each does a dry-run "check-*" and a real "apply-*")
+make check-<service>                   # dry-run, e.g. check-planningalerts, check-righttoknow
+make apply-<service>                   # apply changes, e.g. apply-openaustralia
+STAGE=staging make apply-righttoknow   # scope to a stage/host group (righttoknow, openaustralia support this)
```
benrfairless: This is wrong in both directions.

`STAGE` isn't optional for righttoknow: `check-righttoknow`/`apply-righttoknow` depend on `stage_required` (`Makefile:250-256`), which is a hard `$(error ...)`.

And openaustralia doesn't support it. `make help` advertises new/old, but `inventory/ec2-hosts` has a bare `[openaustralia]` group with one host and no `openaustralia_new`/`_old`, so `-l openaustralia_new` matches nothing.

Mechanically `$(_STAGE)` (`Makefile:11`) applies to every check/apply target, and `STAGE=all` is special-cased to expand to nothing. Accurate version: righttoknow is the only service with stage sub-groups and the only one where `STAGE` is mandatory.

CLAUDE.md
```diff
@@ -0,0 +1,164 @@
...
+  major upgrades (e.g. Ruby version bumps); everyday deploys still go through Capistrano.
+- **Ansible** (`site.yml`, `roles/`, `group_vars/`, `host_vars/`, `inventory/`) configures the OS and services on
+  top of instances Terraform created: packages, users, MySQL/PostgreSQL, nginx/Apache + certbot, cron, backups,
+  monitoring. `roles/internal/` are OAF-authored roles (one per service, e.g. `righttoknow`, `planningalerts`,
+  `theyvoteforyou`, `mysql`, `base-server`, `oaf.certbot`, `oaf.backup`); `roles/external/` are third-party
+  Galaxy roles installed by `make requirements`/`make roles` (not linted, since we don't control their layout).
+- Inventory and `group_vars/`/`host_vars/` follow standard Ansible layering — most per-service config lives in
+  `group_vars/<service>.yml` (e.g. `group_vars/righttoknow.yml`, `group_vars/righttoknow_production.yml`/`_staging.yml`
+  for stage overrides). `group_vars/all.yml` holds cross-service defaults (backup settings, `github_users` allowed
+  to SSH in, etc.). `inventory/ec2-hosts` is the static inventory actually in use — every Makefile target passes it
```
benrfairless: "every Makefile target passes it explicitly" isn't true. `all`, `letsencrypt`, `retry` and `update-github-ssh-keys` pass no `-i` and fall back to `inventory = inventory` in `ansible.cfg:31`, which is the whole directory rather than just `ec2-hosts`.

The conclusion that `aws_ec2.yml` isn't live still holds, and there's a stronger reason for it: `amazon.aws` isn't in `roles/requirements.yml` (only `community.postgresql`), so the plugin can't load at all. Worth saying it that way, since the current wording implies `make all` and `make apply-*` see the same host set.

CLAUDE.md
```diff
@@ -0,0 +1,164 @@
...
+make show-vars HOST=<host>             # dump all Ansible vars for a host
+make show-facts HOST=<host>            # dump all Ansible facts for a host
+make letsencrypt                       # force-renew all registered LetsEncrypt certs
+
+# Terraform (always run tf-secrets/tf-env-check first — the tf-* targets depend on them)
+make tf-plan / make tf-apply           # plan/apply the whole terraform/ config
+make tf-plan-target TARGET=<module>    # scope to one module, e.g. TARGET=planningalerts
+make tf-validate                       # tf-check-fmt + terraform validate
+
+make vagrant                           # bring up local Vagrant VMs for testing Ansible roles (not app dev)
```
benrfairless: `make vagrant` doesn't bring up VMs. `Makefile:140` is `vagrant: .make/vagrant-plugins .make/certificates requirements`, so it installs the `vagrant-hostsupdater` plugin, generates the dev `*.test` certificates and runs requirements. `vagrant up <box>` is the separate step. `make help` has this one right.

CLAUDE.md
```diff
@@ -0,0 +1,164 @@
...
+
+- Needs Python 3.11 (not 3.12+, which breaks Ansible 2.9/2.10 — see `.python-version`). `mise install` sets up
+  Ruby/Python/PHP versions.
+- Ansible Vault passphrases and AWS/Terraform state creds come from the OAF 1Password account (`op signin --account
+  oaforgau`), fetched via `bin/ansible-vault-client`. AWS and Google credentials are your own CLI config (`aws
+  configure`, `gcloud auth application-default login`), not stored here.
+- `make requirements` — sets up the Python venv, installs ansible-galaxy roles/collections, checks 1Password.
+- `make tf-secrets` — renders `terraform/secrets.auto.tfvars` from 1Password (RDS admin password, Cloudflare/Linode
+  API tokens) before any `tf-*` target.
+- `make tf-env-check` — verifies AWS/Cloudflare/Linode/gcloud credentials are reachable.
```
benrfairless: `tf-env-check` doesn't check Cloudflare or Linode, and doesn't fail. `Makefile:134-138` probes exactly two things, `aws sts get-caller-identity` and `gcloud auth application-default print-access-token`, and only prints `WARN:` (the comment above it says "Non-fatal").

Cloudflare and Linode tokens come from `make tf-secrets` now and aren't per-operator, as the README credentials section says. The "AWS/Cloudflare/Linode/gcloud" wording is stale in `Makefile:40` too.

CLAUDE.md
```diff
@@ -0,0 +1,164 @@
...
+STAGE=staging make apply-righttoknow   # scope to a stage/host group (righttoknow, openaustralia support this)
+TAGS=foo SKIP_TAGS=bar make apply-X    # limit/skip Ansible tags
+ANSIBLE_VERBOSE=vvv make apply-X       # verbose ansible-playbook output
+make all                               # run site.yml against every host
+make retry                             # re-run site.yml limited to hosts that failed last time
+make show-vars HOST=<host>             # dump all Ansible vars for a host
+make show-facts HOST=<host>            # dump all Ansible facts for a host
+make letsencrypt                       # force-renew all registered LetsEncrypt certs
+
+# Terraform (always run tf-secrets/tf-env-check first — the tf-* targets depend on them)
```
benrfairless: Not quite true for two of the targets in this block: `tf-validate` and `tf-check-fmt` (`Makefile:228,231`) depend on neither `tf-secrets` nor `tf-env-check`. Useful to spell out, because it means the Terraform half of `make lint` works for someone with no 1Password or AWS access.

benrfairless (commented):
Went through this line by line against the Makefile, `ansible.cfg` and `inventory/ec2-hosts`. The structure and the vault/tag/Terraform sections all check out. There are a handful of places where the commands don't match what the Makefile actually does, and one of them is a safety gap worth fixing before this lands.

These all apply to whichever file holds the content after the AGENTS.md rename, so probably worth doing the rename first and then applying them.

One thing worth fixing at the source: three of the comments below trace back to `make help` being stale (the openaustralia new/old stages, and the Cloudflare/Linode wording in tf-env-check). Correcting those two help lines in this PR too would stop it drifting back next time someone writes docs by reading `make help`.

Reviewed with assistance from Claude Code (claude-opus-5).

CLAUDE.md
```diff
@@ -0,0 +1,164 @@
...
+- If you only have access to some of the four vault IDs (`default`, `all`, `ec2`, `rtk`), set
+  `ANSIBLE_VAULT_IDENTITY_LIST` in `.envrc` listing only the ones you can read (see README "Access to everything
+  except right to know").
+
+## Common commands
+
+```
+make help                              # full list of targets with descriptions
+make lint                              # yaml-lint + template-check + ansible-lint + tf-check-fmt + tf-validate
+make yaml-lint                         # yamllint on roles/ and site.yml only
```
benrfairless: Both `yaml-lint` and `ansible-lint` run against `roles/internal/ roles/*.yml site.yml` (`Makefile:324,337`), not `roles/`.

While you're in the lint section: `.ansible-lint` also excludes `roles/internal/righttoknow/files/storage.yml`, not just `roles/external`. Worth a mention, since the file looks like Ansible YAML but isn't, and the existing comment explaining why is a good one.

CLAUDE.md
```diff
@@ -0,0 +1,164 @@
...
+
+This repo is worked on via terminal Claude Code and other editors alike — keep guidance and commands in this file
+tool-agnostic (plain shell/`make` commands, not editor-specific steps), both when following it and when adding to
+it.
+
+## Setup (one-time, per operator)
+
+- Needs Python 3.11 (not 3.12+, which breaks Ansible 2.9/2.10 — see `.python-version`). `mise install` sets up
+  Ruby/Python/PHP versions.
+- Ansible Vault passphrases and AWS/Terraform state creds come from the OAF 1Password account (`op signin --account
```
benrfairless: Two small ones here.

`op-check` reads the account from `bin/.op-account` (`Makefile:105-117`) rather than a hardcoded value. It contains `oaforgau` today so this is correct, but pointing at the file removes a drift risk.

And a question on `aws configure` below: the README says "whichever AWS auth method we're currently using (`aws configure sso`, `aws configure`, etc.)", `tf-env-check` suggests `aws sso login` in its warning, and `terraform/providers.tf:2` lists `AWS_PROFILE` among the options. If there's a specific profile or SSO flow Terraform needs here, that's exactly the kind of gotcha this file is for and worth naming.

ianheggie-oaf: @benrfairless ,
RE README - I will in a later PR update to read as follows, but since the change has yet to occur, I wont include it here.

 - AWS CLI (aws) — required for Terraform's AWS provider and for reading S3-backed Terraform state. Sign in with aws login (not aws configure sso/AWS SSO); Terraform reads through the oaf-legacy profile, which bridges to that session via credential_process for tools that don't understand login_session yet — set AWS_PROFILE=oaf-legacy (e.g. in .envrc) or pass --profile oaf-legacy per command. Ansible doesn't use this at all: it reaches hosts over SSH, and the handful of tasks that call AWS APIs directly (RDS facts) use their own Ansible Vault-stored access key, independent of your local AWS session.

Thanks for the thorough line-by-line review, this was really useful.

All eight inline corrections are applied:

- STAGE is now documented as mandatory for righttoknow only (staging/production/all), and openaustralia's non-existent new/old groups are gone from both AGENTS.md and the two stale `make help` lines.
- tf-env-check now correctly says AWS/gcloud only, non-fatal warning, no Cloudflare/Linode.
- tf-validate/tf-check-fmt noted as needing neither tf-secrets nor tf-env-check.
- yaml-lint/ansible-lint scope corrected to `roles/internal/`, `roles/*.yml`, `site.yml` (in AGENTS.md and the matching stale `make help` lines), plus a note on the storage.yml exclusion.
- vagrant corrected to "installs the plugin/certs/requirements", not "brings up VMs".
- The inventory paragraph now scopes the `-i` claim to check-*/apply-* only, and gives the real reason aws_ec2.yml can't load (amazon.aws missing from roles/requirements.yml).
- op-check now points at bin/.op-account instead of a hardcoded value.

On the blocking comment: rather than just adding `all`, `retry` and `update-github-ssh-keys` to the dangerous-commands list, I fixed the actual gap. All four of `all`, `retry`, `update-github-ssh-keys` and `letsencrypt` are now wrapped with `bin/tag-provisioning`, so a failed run of any of them leaves the same `wip-*` tag marker `apply-*`/`tf-apply*` already have. The "which commands are live-fire" guidance is also reworked to default-deny: an explicit safe/read-only list, everything else needs a go-ahead, so a future missing target can't silently slip through the way these three did.

On the AWS profile question: for now AGENTS.md just says no specific profile is required today. Pinning to `oaf-legacy` is separate work tracked on #558, not part of this PR.

Pushed to the branch, ready for another look.

This reply and the fixes above were drafted with Claude Code (claude-sonnet-5) assistance, reviewed by me before pushing.

benrfairless (approved):

Makefile
```diff
@@ -172,13 +172,19 @@ venv: .venv/bin/activate
 roles: .make/roles
 
 all: requirements
+	bin/tag-provisioning --wip all "" "" ""
 	.venv/bin/ansible-playbook site.yml
+	bin/tag-provisioning all "" "" ""
```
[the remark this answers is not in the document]
ianheggie-oaf: I don't regard this as a problem

benrfairless (approved):

PR #629, opened 2026-08-10
Record emergency AWS changes in terraform

benrfairless:
## Description

Records in Terraform a set of changes that were made directly in the AWS console during an emergency, so that config and live infrastructure describe the same thing again.

The console changes were four ALB listener rules that block ten source IP ranges with a `403 Forbidden`, sitting at priorities 1-4 on the HTTPS listener. Those are added here in `terraform/blocked-source-ips.tf`, with `import` blocks so Terraform adopts the existing rules rather than creating duplicates.

Inserting them pushed every managed listener rule down, so four priorities are corrected to match live: `redirect-https-to-canonical` 1 to 10, `main-https-redirect-sitemaps` 2 to 11, `forward` 5 to 12, and metabase's `main` 6 to 13.

Two smaller pieces of drift are also picked up. `maindb` is running MySQL 8.4.8 while config pinned 8.4.6, because `auto_minor_version_upgrade` is enabled and AWS moves the patch version by itself; config now tracks the `8.4` family instead, which is the AWS provider's [documented way](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance#engine_version) to handle exactly this and stops it re-drifting at every patch. The RightToKnow staging instance had also gained an `AnsibleGroup` tag out of band, which is now in config.

## Motivation and Context

Terraform was out of sync with production after emergency console work, which means a plan could not be trusted and the next apply risked reverting the IP blocks.

A full sweep was run to find everything, not just the known changes: a `terraform plan -refresh-only` for attribute drift on managed resources, a normal plan for what Terraform would revert, and a read-only comparison of live AWS and Cloudflare API listings against state to catch resources that exist but were never in Terraform at all.

Most of the estate came back clean. All 14 security groups match state with no rule drift, all 173 Terraform-managed DNS records match Cloudflare exactly with nothing deleted, and EC2 instances, target groups, ACM certificates and the `us-east-1`/`ap-southeast-1` regions are all as described. The blast radius really was just the load balancer.

Two unrelated findings came out of the sweep and are deliberately **not** changed here, to keep this reviewable:

- #627 — the plan carries six destroys queued by earlier cleanup commits (23a8b8fc, b5cda440) that were committed but never applied. Worth reading before anyone applies this branch, since the same apply executes them.
- #628 — eight Cloudflare DNS records and three idle Elastic IPs exist in production but are not managed by Terraform.

## How Has This Been Tested?

Verified with a read-only `terraform plan` against production. Every difference attributable to the console changes is gone, and the four blocking rules now show as `will be imported` with no accompanying diff, which confirms the config matches the live rules exactly.

`make tf-check-fmt` and `make tf-validate` both pass locally.

No `terraform apply` has been run. State does not yet contain the imports, so applying this branch is still to be done deliberately.

Remaining plan output is expected and unrelated: the `aws_key_pair.deployer` replacement documented in README, plus the six destroys tracked in #627.

## Types of Changes

- [ ] Documentation

## Checklist:

## Notes for the reviewer

The branch name has no issue number, which departs from Conventional Branch. There was no tracked issue for this reconciliation work; #627 and #628 are follow-ups filed from it rather than its parent. Happy to renumber if you would prefer an issue raised retrospectively.

Worth a look while you are in here: `terraform/security-groups.tf` declares the VPN SSH rule as an inline `ingress` block on the `webserver` and `planningalerts` groups, while `terraform/vpn-security-groups.tf` declares the same rule again as standalone `aws_security_group_rule` resources. Terraform treats inline blocks as authoritative for the whole group, so the two representations can fight each other. It is pre-existing and did not cause drift here, so it is left alone, but it is a trap for later.

Assisted-by: Claude Code/claude-opus-5

### DIFF
--- a/terraform/blocked-source-ips.tf
+++ b/terraform/blocked-source-ips.tf
@@ -1,0 +1,131 @@
+# Source IP ranges blocked at the load balancer with a 403.
+#
+# These rules were added directly in the AWS console as an emergency response and
+# are recorded here so they're explicit, reviewable, and removed deliberately rather
+# than by accident. They sit at priorities 1-4, ahead of every routing rule on the
+# HTTPS listener, so blocked traffic never reaches an application.
+#
+# The grouping of CIDRs across the four rules matches what was created in the console.
+# Each ALB rule condition allows a limited number of values, so ranges were added in
+# batches as they were identified.
+#
+# The `import` blocks below adopt the existing rules into Terraform state rather than
+# creating new ones. They are a no-op once applied and can be deleted after that.
+
+resource "aws_lb_listener_rule" "blocked_source_ips_1" {
+  listener_arn = aws_lb_listener.main-https.arn
+  priority     = 1
+
+  action {
+    type = "fixed-response"
+
+    fixed_response {
+      content_type = "text/plain"
+      message_body = "Forbidden"
+      status_code  = "403"
+    }
+  }
+
+  condition {
+    source_ip {
+      values = [
+        "43.173.176.0/22",
+        "43.173.180.0/23",
+        "43.173.182.0/24",
+      ]
+    }
+  }
+}
+
+resource "aws_lb_listener_rule" "blocked_source_ips_2" {
+  listener_arn = aws_lb_listener.main-https.arn
+  priority     = 2
+
+  action {
+    type = "fixed-response"
+
+    fixed_response {
+      content_type = "text/plain"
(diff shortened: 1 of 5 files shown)


ianheggie-oaf (approved):
Approved - LGTM and explains the changes I noticed (The sentry suggestion makes sense, I presume you will do it)

PR #643, opened 2026-08-12
Ansible over SSM configured for all hosts except vpn/old server and added extra tools

ianheggie-oaf:
## Summary

- Originally scoped to righttoknow-staging's SSM access; repurposed mid-branch to move (almost) the
  whole EC2 fleet onto Ansible's dynamic `aws_ec2` inventory - updated:
  - theyvoteforyou, 
  - openaustralia  (production), 
  - proxy, 
  - metabase, 
  - righttoknow (production + staging),
  - planningalerts. 
- Left out of the migration to the dynamic inventory (both since removed later in this stack):
    - The VPN server - stays on the static `inventory/ec2-hosts` entry, though it did pick up the same
      `PublicHostname`/`LogName` tags for consistency (`terraform/vpn-server.tf`); the server itself is
      fully removed in #679.
    - The already-stopped legacy "OLD openaustralia" instance (`terraform/openaustralia/main.tf`) - it
      never had an `AnsibleGroups` tag, so it was never part of the Ansible-managed fleet either way;
      removed in #650.
- New `PublicHostname`/`LogName` EC2 tags and matching `public_hostname`/`log_name` Ansible vars
  (`group_vars/all.yml`/`ssm.yml`) - needed because the dynamic inventory's hostname is no longer a
  stable identifier across an instance replacement. 
    - `public_hostname` is deliberately shared across a blue/green fleet (backup paths etc.);
    -  `log_name` is deliberately unique per instance (CloudWatch log streams and the name displayed by ansible logs (`inventory_hostname`)).
- New operational tooling: `make ssh-config` (generates an SSM-proxied `~/.ssh/config` block) and
  `make server-status [HOST=...]` (fleet-wide uptime/memory/disk/failed-units check), plus
  `bin/show-aws-inventory` updated to show `LogName` as the primary column.
- Did **not** implement locking sshd to localhost, so we can still ssh between servers if SSM agent ever stops working (as happened when snapd was removed a few weeks ago from one server to minimise memory use)
  - Will rely on future PR to change security rules to block public ssh access
- Pinned `amazon.aws`/`community.aws`/`community.postgresql` collection versions so the dynamic
  inventory plugin loads under our `ansible-core` pin, which forced migrating the now-removed `rds`
  module to `rds_instance_info` (same underlying value, different param/return names).

## Motivation

Groundwork for #558 (move fleet access onto SSM/IAM-with-MFA, retiring direct SSH) - also unblocks
the Cloudflare Tunnel work (oaf-internal#266), which needs SSM as the primary access path before
direct SSH access changes further.

Originally just "get righttoknow-staging working over SSM"; repurposed, per Ian, to cover the whole
fleet in one pass rather than repeating the same tag/inventory work incrementally per service.

## Note for reviewers/ops: planningalerts CloudWatch log streams

Unlike everywhere else, planningalerts' `LogName` mirrors each instance's `Name` tag rather than
its old EC2-assigned hostname (see `terraform/planningalerts/env/main.tf`'s comment for why - its
instances get replaced too often for a frozen snapshot to stay meaningful). Since this tag has now
been applied for real, planningalerts' CloudWatch log streams (`/var/log/syslog`,
`/var/log/unattended-upgrades/unattended-upgrades.log`, `.../production.log`, `.../puma.log`) have
started new streams under the new names; the old streams stop receiving data but keep their history
(CloudWatch doesn't rename in place). Mapping at the time this was applied:

- `ec2-3-27-189-101.ap-southeast-2.compute.amazonaws.com` → `web1.blue.planningalerts`
- `ec2-3-27-58-54.ap-southeast-2.compute.amazonaws.com` → `web2.blue.planningalerts`

## AI involvement

Drafted with Claude Code (claude-sonnet-5) - see the `Assisted-by` trailers on the commits for the
full disclosure.

## Test plan

- [x] `make tf-check-fmt` / `make tf-validate` / full `make tf-plan` pass; tags applied for real via
      `tf-apply` (fleet now carries `PublicHostname`/`LogName`/`AnsibleGroups`).
- [x] `inventory/aws_ec2.yml` YAML parses (`make yaml-lint` doesn't cover `inventory/` - checked
      manually via `yaml.safe_load`).
- [x] `make ssh-config` tested against real AWS.
- [x] `make ssh-config` re-tested since pinning it to `--profile oaf` (last run surfaced an expired
      browser session, not yet re-verified clean).
- [x] `make server-status`  tested against real AWS.
- [x] Confirm SSH from another OAF-managed instance on the same VPC still works as a fallback
      (carried over from this branch's original scope).
      Checked by `ssh -A theyvoteforyou`, looking up private IP on ec2 console for righttoknow-staging, and then `ssh root@172.31.34.95`)
- [x] Run `make check-<service>` against all hosts now on the dynamic inventory
      with the new `log_name`/`public_hostname` vars, to confirm nothing regressed.
- [x] Run `make check-<service>`/`apply-<service>` against a real host now on the dynamic inventory
      with the new `log_name`/`public_hostname` vars, to confirm nothing regressed - not yet done
      for any service besides righttoknow-staging (which predates these latest changes).
```
$ make server-status
OK: OAF 1Password reachable (oaforgau)
OK: recent aws CLI found (aws-cli/2.36.2 Python/3.14.6 Linux/6.17.0-19-generic exe/x86_64.ubuntu.24)
OK: session-manager-plugin found (1.2.835.0)
OK: [profile oaf] already has a login_session set
OK: [profile oaf-legacy] already configured
bin/server-status
srv.theyvoteforyou.org.au | CHANGED | rc=0 >>
ip-172-31-3-170
Thu Aug 13 15:22:58 AEST 2026

 15:22:58 up 1 day, 15:54,  1 user,  load average: 0.27, 0.26, 0.27

              total        used        free      shared  buff/cache   available
Mem:          15803        1775        1092           0       12935       13687
Swap:             0           0           0

Filesystem     1M-blocks  Used Available Use% Mounted on
/dev/root          15807  7917      7875  51% /
/dev/nvme1n1       30130 15300     13279  54% /data

Filesystem      Inodes  IUsed   IFree IUse% Mounted on
/dev/root      2048000 302263 1745737   15% /
/dev/nvme1n1   1966080  87181 1878899    5% /data

Failed units:
certbot.service       loaded failed failed Certbot                               
fwupd-refresh.service loaded failed failed Refresh fwupd metadata and update motd
================================================================================
(etc for rest of servers, I have listed just the failed services below)
================================================================================
openaustralia.org.au | CHANGED | rc=0 >>
...
Failed units:
* certbot.service loaded failed failed Certbot
================================================================================
prod.righttoknow.org.au | CHANGED | rc=0 >>
...
Failed units:
* varnishncsa.service loaded failed failed Varnish HTTP accelerator log daemon
================================================================================
```

All three failed units above (`certbot.service`, `fwupd-refresh.service`, `varnishncsa.service`) are
pre-existing on their hosts, not caused by this PR - `make server-status` is simply the first tool that
surfaced them fleet-wide. Filed as #648 for separate investigation rather than blocking this PR on them.

## TODO In later PR

*  Confirm `sshd` on an `ssm`-group host still refuses external connections once the relevant
      security group is tightened (follow-up, not yet done - carried over from this branch's
      original scope).

## Types of Changes

- [ ] Documentation

## Checklist:

- [x] My change requires a change to the documentation.


bin/ssh-config
```diff
@@ -0,0 +1,60 @@
...
+aws ec2 describe-instances --profile oaf \
+  --filters "Name=tag-key,Values=PublicHostname" "Name=instance-state-name,Values=running" \
+  --query "Reservations[].Instances[].{Id:InstanceId,Name:Tags[?Key=='Name']|[0].Value,PublicHostname:Tags[?Key=='PublicHostname']|[0].Value}" \
+  --output json |
+  jq -r --arg identity_file "$identity_file" \
+    '[.[] | select(.PublicHostname != null and .PublicHostname != "")] | sort_by(.PublicHostname) | .[] |
+    "Host \(.Name) \(.PublicHostname) \(.Id)\n" +
+    "    ProxyCommand sh -c \"aws ssm start-session --profile oaf --target \(.Id) --document-name AWS-StartSSHSession --parameters portNumber=%p\"\n" +
+    "    User root\n" +
+    (if $identity_file != "" then "    # IdentityFile \($identity_file)\n    # IdentitiesOnly yes\n" else "" end)'
```
benrfairless: **must:** This emits `PublicHostname` as a `Host` alias on every instance, but `PublicHostname` is *by design* identical across a fleet. For planningalerts that means `www.planningalerts.org.au` appears as an alias in two `Host` blocks (four if blue and green are both up). OpenSSH takes the first matching block, so `ssh www.planningalerts.org.au` lands on an arbitrary instance with no warning: you think you're on web1, you're on web2.

Suggestion below emits the `PublicHostname` alias only where it's unique across the result set, so the shared ones fall away and each host keeps its unique `Name` and instance ID. I tested the jq against fixture data with two colliding planningalerts entries plus unique metabase/righttoknow ones and it behaves as intended.

```suggestion
    '[.[] | select(.PublicHostname != null and .PublicHostname != "")]
    | (group_by(.PublicHostname) | map(select(length == 1) | .[0].PublicHostname)) as $unique
    | sort_by(.PublicHostname) | .[] | . as $i |
    "Host \(.Name)\(if ($unique | index($i.PublicHostname)) then " \(.PublicHostname)" else "" end) \(.Id)\n" +
    "    ProxyCommand sh -c \"aws ssm start-session --profile oaf --target \(.Id) --document-name AWS-StartSSHSession --parameters portNumber=%p\"\n" +
    "    User root\n" +
    (if $identity_file != "" then "    # IdentityFile \($identity_file)\n    # IdentitiesOnly yes\n" else "" end)'
```

inventory/aws_ec2.yml
```diff
@@ -2,7 +2,45 @@ plugin: aws_ec2
...
+  # inventory_hostname, since that's no longer the same value. This is also what %h resolves to
+  # below, and what your ~/.ssh/config `Host i-*` block (SSH key etc.) matches against.
+  ansible_host: instance_id
+  # SSH over SSM, not a direct network connection - same for every host here, no per-operator
+  # override possible (or needed): OpenSSH always prefers a command-line -o over ~/.ssh/config for
+  # the same keyword, and this becomes one via ansible_ssh_common_args.
+  # NOTE: compose values are Jinja2 *expressions*, not literal strings - this must be a quoted
+  # Jinja2 string literal (single-quoted here, since the value itself only needs double quotes),
+  # or it silently evaluates to nothing (Ansible's compose failures are non-fatal by default).
+  ansible_ssh_common_args: "'-o ProxyCommand=\"aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters portNumber=%p\"'"
```
benrfairless: **should:** This `aws ssm start-session` has no `--profile`, so it uses whatever's ambient. `bin/ssh-config` pins `--profile oaf` with a whole paragraph explaining it can't rely on `AWS_PROFILE` being set, and that argument applies verbatim here: the ProxyCommand runs at `ssh` time in whatever shell Ansible forks. It works today for anyone who followed either INSTALL.md path (`export AWS_PROFILE=oaf-legacy`, or duplicating the block as `[default]`), but neither is guaranteed, and it's a bit odd that the plugin's own API calls pin `aws_profile: oaf-legacy` two lines up while this one doesn't pin anything.

`oaf` rather than `oaf-legacy` for the same reason `bin/ssh-config` gives: this is a plain `aws` CLI call, which reads `login_session` natively.

(Also adds the missing trailing newline.)

```suggestion
  ansible_ssh_common_args: "'-o ProxyCommand=\"aws ssm start-session --profile oaf --target %h --document-name AWS-StartSSHSession --parameters portNumber=%p\"'"
```

inventory/aws_ec2.yml
```diff
@@ -2,7 +2,45 @@ plugin: aws_ec2
...
+exclude_filters:
+  - tag:AnsibleGroups: [""]
+# Name hosts by their LogName tag, not Name or instance-id/DNS name - but see compose below, which
+# points the actual SSH connection at the instance-id regardless. Name would collide with several
+# static-inventory group names (e.g. the "metabase"/"theyvoteforyou" instances are literally named
+# that, same as the group), and PublicHostname isn't unique enough (deliberately identical across
+# a blue/green fleet - see group_vars/all.yml). LogName is unique per-instance by design and
+# happens to restore each host's original static-inventory name too.
+hostnames:
+  - tag:LogName
```
benrfairless, line 21: **should:** Worth a warning comment here. An instance tagged `AnsibleGroups` but missing `LogName` is silently dropped from the inventory. I confirmed it in the amazon.aws 4.5.0 plugin source: `_add_hosts` does `if not hostname_list: continue`, with no warning, and `strict: true` doesn't help because it covers compose/groups/keyed_groups, not hostname resolution.

The fully-missing case is fine, since `-l <group>` with an empty match is a fatal Ansible error. The one that worries me is *partial*: a multi-host service (planningalerts, or `-l righttoknow` spanning prod and staging) where one host vanishes. The playbook runs on the rest, exits 0, and `bin/tag-provisioning` stamps the success tag, so it reads as a clean run.

There's no plugin option that makes this fatal, so a documented check is about the best available:

```suggestion
hostnames:
  - tag:LogName
# CAUTION: an instance tagged AnsibleGroups but with no LogName tag is silently dropped from
# the inventory - the plugin skips any host it can't name, with no warning, and strict: true
# doesn't cover hostname resolution. Worst case is a partial run: one host of a multi-host
# service disappears, the playbook succeeds on the rest and exits 0. After tagging a new host,
# confirm it shows up in `make show-aws-inventory` before relying on `make apply-<service>`.
```

benrfairless (commented):
## Review: Ansible over SSM for the EC2 fleet

Really solid piece of work, and the inline comments explaining *why* are the best part of it. The collection-pinning notes in `roles/requirements.yml` in particular (the `HAS_BOTO` / `ImportError` explanation for why `community.aws` has to stay version-matched to `amazon.aws`) are exactly the kind of thing that saves the next person hours.

**Blast radius: high.** This changes how every EC2 host is discovered, renames CloudWatch log streams, and moves the restic backup paths. I traced the consumers rather than trusting the diff: there are no remaining `inventory_hostname` uses that carry identity, and the `rds` to `rds_instance_info` migration is complete across every consumer (`group_vars/ec2.yml`, `site.yml`, `packer/playbook.yml`). All 7 CI checks pass.

I've left inline suggestions for the fixable items. Leaving this as a comment rather than "request changes" - the one gating item is a verification run, not a code defect.

### One thing I'd want confirmed before merge

`roles/internal/openaustralia/meta/main.yml:144` and `roles/internal/righttoknow/meta/main.yml:154` now derive the restic repository path from `log_name`, and the test plan marks `apply-*` as not yet run for either service.

I worked through the values and it should be a no-op: `log_name` equals the old `inventory_hostname` exactly for both (`openaustralia.org.au` and `prod.righttoknow.org.au`), which is what 7f311b53 was fixing. But "should be" is the problem. If it isn't, restic initialises a fresh empty repository at a new S3 prefix and backup continuity for production data restarts from zero.

Could you run `make check-openaustralia` and `STAGE=production make check-righttoknow` with `--diff` and confirm the restic profile and cron show no change to the repository path? Both are `check-*`, so read-only, no production go-ahead needed.

### Not inline-able, so noting here

`AGENTS.md`'s known-safe read-only list covers `show-*` (so `show-aws-inventory` is fine) but not `ssh-config` or `server-status`. Per AGENTS.md's own rule, anything off that list gets treated as a live-fire production action needing an explicit go-ahead, which is a bit silly for two read-only helpers. This PR already edits AGENTS.md, so it's a small addition in a file it's touching anyway.

### On the two open Sentry bot comments

Both look wrong to me, flagging so they don't drive a change:

- **`bin/ssh-config:59`, "User root fails, needs ubuntu" (HIGH).** False positive. `roles/internal/deploy-user` authorises `github_users`' personal keys on `root` unrestricted; only the *terraform* key gets the `command="Please login as the user ubuntu"` forced command. `ansible.cfg` sets `remote_user = root` for the whole fleet. `User root` is correct. (One real caveat: a freshly built instance that hasn't been provisioned yet still has cloud-init's restriction, so root fails until the first Ansible run.)
- **`roles/internal/openaustralia/meta/main.yml:144`, "should use public_hostname" (MEDIUM).** The bot has the fix backwards. `public_hostname` is deliberately shared across a fleet, so using it for a restic repository would point two instances at one repo, i.e. concurrent writers on the same backup store. `log_name` is the right choice here. The underlying verification concern is real, and is the gating item above.

### Smaller things, all inline

Alias collision in `bin/ssh-config`, a silent-drop failure mode in the dynamic inventory, a missing `--profile` on the inventory ProxyCommand, `bin/server-status` aborting on a single unreachable host, plus two comment/whitespace nits.

### Things I checked and cleared, so nobody else has to

- `-l <group>` with an empty match is a fatal Ansible error, so a host disappearing entirely from the dynamic inventory fails loudly rather than silently no-opping. It's only the *partial* case that's quiet (see the inline comment on `hostnames`).
- The packer path is fine: `packer/planningalerts.pkr.hcl` sets `inventory_directory = ".."`, so `group_vars/all.yml` is loaded and `log_name` is defined during an AMI build. The `awscloudwatch` template task is gated on `'ec2' in group_names` anyway, and the packer host isn't in `ec2`.
- The proxy host (the Ubuntu 16.04 box that skips the boto3 pip task) isn't in any `requires_*` group, so `db_name` is undefined there and both `rds_instance_info` tasks skip. No boto3 version problem on xenial.
- No host ends up in both `requires_mysql` and `requires_postgresql` under the new `AnsibleGroups` tags, so `site.yml`'s exclusivity assert still holds.

---

*Reviewed with assistance from Claude Code (claude-opus-5). The analysis and inline suggestions were AI-drafted; I've checked them before posting.*

inventory/aws_ec2.yml
```diff
@@ -2,7 +2,45 @@ plugin: aws_ec2
 aws_profile: oaf-legacy
 regions:
   - ap-southeast-2
+# Only discover instances explicitly opted in via a non-blank AnsibleGroups tag (set in
+# Terraform) - deliberately NOT every EC2 instance in the account. Untagged/not-yet-migrated
+# hosts (the untouched rest of the fleet, stray old/stopped instances, etc.) must never appear
+# here at all - membership has to mean "explicitly declared SSM-ready", not "AWS happens to know
+# this instance exists".
+filters:
+  tag-key: AnsibleGroups
```
benrfairless, line 11: **suggestion (optional):** The plugin defaults to `running/pending/stopping/stopped` when no state filter is given, so a stopped-but-tagged instance would turn up as an unreachable inventory host. It's latent today, since a disabled blue/green env is `count = 0` rather than stopped instances, and `bin/ssh-config` already filters `instance-state-name=running`. Mostly this is about the two staying symmetrical.

Happy for this to be a no-op if you'd rather stopped instances stay visible in `make show-aws-inventory`.

```suggestion
filters:
  tag-key: AnsibleGroups
  # The plugin defaults to running/pending/stopping/stopped when no state filter is given -
  # restrict to running so a stopped instance can't appear as an unreachable inventory host
  # (bin/ssh-config already filters the same way).
  instance-state-name: running
```

bin/server-status
```diff
@@ -0,0 +1,23 @@
...
+# of -h and -m). One combined remote command per host (needs -m shell, not command, for the `;`
+# separators) rather than several separate ansible invocations - avoids several times the per-host
+# SSM connection overhead for what's meant to be a quick, pithy check.
+#
+#
+# Any extra arguments are passed straight through to ansible - e.g. --limit planningalerts,
+# --forks 1.
+sep="================================================================================"
+.venv/bin/ansible all -i ./inventory -m shell -a \
+  "hostname; date; echo; uptime; echo; free -m; echo; df -m -x squashfs -x tmpfs -x devtmpfs; echo; df -i -x squashfs -x tmpfs -x devtmpfs; echo; echo 'Failed units:'; systemctl --failed --no-legend ; echo $sep" "$@"
```
benrfairless: **should:** `set -euo pipefail` plus the ERR trap means one unreachable host turns the whole fleet health check into `ERROR: server-status failed at line 22`, with the useful per-host output scrolled off above it. That'll fire for most people: `ansible all` picks up the static `[openvpn] 52.64.33.12` entry keyed to `terraform.pem`, on a box slated for retirement (#558).

For a status command I think reporting and carrying on beats aborting:

```suggestion
sep="================================================================================"
rc=0
.venv/bin/ansible all -i ./inventory -m shell -a \
  "hostname; date; echo; uptime; echo; free -m; echo; df -m -x squashfs -x tmpfs -x devtmpfs; echo; df -i -x squashfs -x tmpfs -x devtmpfs; echo; echo 'Failed units:'; systemctl --failed --no-legend ; echo $sep" "$@" || rc=$?
if [ "$rc" -ne 0 ]; then
  echo "WARNING: ansible exited $rc - one or more hosts were unreachable or failed (see per-host output above)" >&2
fi
```

terraform/righttoknow/staging.tf
```diff
@@ -7,10 +7,18 @@ resource "aws_instance" "staging" {
...
-    Purpose      = "Ubuntu 22.04 Staging Server"
-    AnsibleGroup = "righttoknow_staging"
+    Name        = "righttoknow-staging"
+    Environment = "staging"
+    Purpose     = "Ubuntu 22.04 Staging Server"
+    # Flattened inventory/ec2-hosts group membership, including groups only reached via
+    # :children (righttoknow, catch_all_mail) - see inventory/aws_ec2.yml.
+    AnsibleGroups = "ec2,righttoknow,righttoknow_staging,catch_all_mail,requires_postgresql"
+    # Matches the exact old static-inventory hostname, so migrating to the dynamic inventory
+    # doesn't move backup_hostname (see group_vars/all.yml/ssm.yml) and orphan backup history.
```
benrfairless: **nit:** `backup_hostname` doesn't exist anywhere in the repo. The backup path continuity is also carried by `LogName` rather than `PublicHostname`, since the restic repository is derived from `log_name` - so as written the rationale is attached to the wrong tag.

```suggestion
    # Matches the exact old static-inventory hostname. (Backup continuity is actually carried
    # by LogName below - the restic repository path is derived from log_name, not
    # public_hostname; see roles/internal/righttoknow/meta/main.yml.)
```

roles/internal/deploy-user/tasks/main.yml
```diff
@@ -71,3 +71,4 @@
     line: "PasswordAuthentication no"
   when: add_authorized_keys is success
   notify: sshd restart
+
```
benrfairless: **nit:** Looks like a leftover from reverting the lock-sshd-to-127.0.0.1 change - this is the only thing left in the file's diff.

```suggestion
```

roles/internal/openaustralia/meta/main.yml
```diff
@@ -134,14 +134,14 @@ dependencies:
...
-  #       target: "s3://s3.amazonaws.com/oaf-backups/openaustralia/{{ inventory_hostname }}/data"
+  #       target: "s3://s3.amazonaws.com/oaf-backups/openaustralia/{{ log_name }}/data"
 
   - role: oaf.restic
     restic_profiles:
       - name: openaustralia_data
         schedule: "0 3 * * *"
         source: /data
-        repository: "s3:s3.amazonaws.com/oaf-backups/openaustralia/{{ inventory_hostname }}/restic"
+        repository: "s3:s3.amazonaws.com/oaf-backups/openaustralia/{{ log_name }}/restic"
```
benrfairless, line 144: **must (verify, not change):** This is the gating item from my review body. `log_name` should resolve to `openaustralia.org.au` here, identical to the old `inventory_hostname`, so the repository path shouldn't move. Worth confirming with `make check-openaustralia --diff` before merge though, because if it does move restic will init a fresh empty repo at the new prefix and backup history for production data effectively restarts.

Same check for `roles/internal/righttoknow/meta/main.yml:154` via `STAGE=production make check-righttoknow`.

On the Sentry bot's comment on this line: I think it's backwards. `public_hostname` is deliberately shared across a fleet, so using it for a restic repository would point every instance in a pool at the same repo, i.e. concurrent writers on one backup store. `log_name` is the correct choice.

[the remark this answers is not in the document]
ianheggie-oaf: The actual s3 path is s3://oaf-backups/openaustralia/openaustralia.org.au/restic/

`group_vars/all.yml` actually says to use `log_file`: 
> ... use public_hostname (not unique) or log_name
  (unique) instead, e.g. for backup paths (see roles/internal/righttoknow's restic repository)

Given terraform/openaustralia/production.tf` has
```
    PublicHostname = "www.openaustralia.org.au"
    LogName        = "openaustralia.org.au"
```

Then LogFile IS the correct choice!

bin/ssh-config
```diff
@@ -0,0 +1,60 @@
...
+
+aws ec2 describe-instances --profile oaf \
+  --filters "Name=tag-key,Values=PublicHostname" "Name=instance-state-name,Values=running" \
+  --query "Reservations[].Instances[].{Id:InstanceId,Name:Tags[?Key=='Name']|[0].Value,PublicHostname:Tags[?Key=='PublicHostname']|[0].Value}" \
+  --output json |
+  jq -r --arg identity_file "$identity_file" \
+    '[.[] | select(.PublicHostname != null and .PublicHostname != "")] | sort_by(.PublicHostname) | .[] |
+    "Host \(.Name) \(.PublicHostname) \(.Id)\n" +
+    "    ProxyCommand sh -c \"aws ssm start-session --profile oaf --target \(.Id) --document-name AWS-StartSSHSession --parameters portNumber=%p\"\n" +
+    "    User root\n" +
```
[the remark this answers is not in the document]
ianheggie-oaf: I have updated ssh-config to set `deploy` as the default `User` for better security.

Note: the ansible playbook actually enables ssh to root, so this advice was incorrect. for example:
```
ianh@ben:.../infrastructure (feature/ansible-via-s)$ ssh web1.blue.planningalerts id
bash: warning: setlocale: LC_ALL: cannot change locale (en_AU.UTF-8)
uid=0(root) gid=0(root) groups=0(root)
ianh@ben:.../infrastructure (feature/ansible-via-s)$ ssh ubuntu@web1.blue.planningalerts id
bash: warning: setlocale: LC_ALL: cannot change locale (en_AU.UTF-8)
uid=1000(ubuntu) gid=1000(ubuntu) groups=1000(ubuntu),4(adm),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),119(netdev),120(lxd)
ianh@ben:.../infrastructure (feature/ansible-via-s)$ ssh deploy@web1.blue.planningalerts id
bash: warning: setlocale: LC_ALL: cannot change locale (en_AU.UTF-8)
uid=1001(deploy) gid=1001(deploy) groups=1001(deploy)
```

Never the less,

inventory/aws_ec2.yml
ianheggie-oaf: @benrfairless This makes it silently ignore pending/stopping/stopped rather than complain. I would prefer it to complain based on "Least surprise principle".

When I run check-* or apply-* I expect it to complain if it cannot do as asked!

I have rejected the suggestion, though if you feel strongly about it, lets discuss it.

roles/internal/openaustralia/meta/main.yml
ianheggie-oaf: Has already been updated

bin/server-status
ianheggie-oaf: It now reports and carries on, AND it checks the commands ran (reported as a failure and shows debug output on non zero output).

Treats presence of failed units as "changed" (orange text), No failed units as OK (green), and reports down systems, eg when I stopped openvpn, it reported:

```
$->(1) bin/server-status 

PLAY [all] ************************************************************************************************************************************************************************************************

TASK [Collect status] *************************************************************************************************************************************************************************************
changed: [srv.theyvoteforyou.org.au]
changed: [staging.righttoknow.org.au]
changed: [openaustralia.org.au]
changed: [prod.righttoknow.org.au]
ok: [web1.blue.planningalerts]
ok: [web2.blue.planningalerts]
ok: [au.proxy.oaf.org.au]
ok: [web.metabase.oaf.org.au]
fatal: [52.64.33.12]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ssh: connect to host 52.64.33.12 port 22: Connection timed out", "unreachable": true}

TASK [Show status] ****************************************************************************************************************************************************************************************
changed: [openaustralia.org.au] => {
    "status_result.stdout_lines": [
        "openaustralia-org-au",
        "Sun Aug 16 01:04:17 AEST 2026",
        "",
        " 01:04:17 up 4 days,  2:34,  1 user,  load average: 0.06, 0.08, 0.08",
        "",
        "               total        used        free      shared  buff/cache   available",
        "Mem:            1930         331         175          14        1423        1392",
        "Swap:           1023          64         959",
        "",
        "Filesystem      1M-blocks  Used Available Use% Mounted on",
        "/dev/root           19663 10980      8667  56% /",
        "/dev/nvme1n1        19987 12724      6224  68% /data",
        "/dev/nvme0n1p15       105     7        99   6% /boot/efi",
        "",
        "Filesystem       Inodes  IUsed   IFree IUse% Mounted on",
        "/dev/root       2580480 364400 2216080   15% /",
        "/dev/nvme1n1    1310720  17226 1293494    2% /data",
        "/dev/nvme0n1p15       0      0       0     - /boot/efi",
        "",
        "Failed units:",
        "* certbot.service loaded failed failed Certbot"
    ]
}
changed: [srv.theyvoteforyou.org.au] => {
    "status_result.stdout_lines": [
        "ip-172-31-3-170",
        "Sun Aug 16 01:04:17 AEST 2026",
        "",
        " 01:04:17 up 4 days,  1:36,  1 user,  load average: 0.18, 0.23, 0.26",
        "",
        "              total        used        free      shared  buff/cache   available",
        "Mem:          15803        1612        1048           0       13142       13850",
        "Swap:             0           0           0",
        "",
        "Filesystem     1M-blocks  Used Available Use% Mounted on",
        "/dev/root          15807  8031      7761  51% /",
        "/dev/nvme1n1       30130 15300     13279  54% /data",
        "",
        "Filesystem      Inodes  IUsed   IFree IUse% Mounted on",
        "/dev/root      2048000 317417 1730583   16% /",
        "/dev/nvme1n1   1966080  87181 1878899    5% /data",
        "",
        "Failed units:",
        "certbot.service       loaded failed failed Certbot                               ",
        "fwupd-refresh.service loaded failed failed Refresh fwupd metadata and update motd"
    ]
}
changed: [prod.righttoknow.org.au] => {
    "status_result.stdout_lines": [
        "ip-172-31-38-219",
        "Sun Aug 16 01:04:17 AEST 2026",
        "",
        " 01:04:17 up 4 days,  2:34,  1 user,  load average: 0.65, 0.41, 0.28",
        "",
        "               total        used        free      shared  buff/cache   available",
        "Mem:            7840        1720         413           3        5706        5795",
        "Swap:           4607         117        4490",
        "",
        "Filesystem      1M-blocks   Used Available Use% Mounted on",
        "/dev/root           19663  11196      8452  57% /",
        "/dev/nvme1n1       240820 131756     96761  58% /data",
        "/dev/nvme0n1p15       105      7        99   6% /boot/efi",
        "",
        "Filesystem        Inodes   IUsed    IFree IUse% Mounted on",
        "/dev/root        2580480  292297  2288183   12% /",
        "/dev/nvme1n1    15728640 3042113 12686527   20% /data",
        "/dev/nvme0n1p15        0       0        0     - /boot/efi",
        "",
        "Failed units:",
        "* varnishncsa.service loaded failed failed Varnish HTTP accelerator log daemon"
    ]
}
ok: [web1.blue.planningalerts] => {
    "status_result.stdout_lines": [
        "ip-172-31-2-63",
        "Sun Aug 16 01:04:17 AEST 2026",
        "",
        " 01:04:17 up 4 days,  2:34,  1 user,  load average: 0.24, 0.20, 0.18",
        "",
        "               total        used        free      shared  buff/cache   available",
        "Mem:            3836        2222         225           2        1388        1325",
        "Swap:            511          27         484",
        "",
        "Filesystem      1M-blocks  Used Available Use% Mounted on",
        "/dev/root           15693  8890      6788  57% /",
        "efivarfs                1     1         1   4% /sys/firmware/efi/efivars",
        "/dev/nvme0n1p15       105     7        99   6% /boot/efi",
        "",
        "Filesystem       Inodes  IUsed   IFree IUse% Mounted on",
        "/dev/root       2064384 222932 1841452   11% /",
        "efivarfs              0      0       0     - /sys/firmware/efi/efivars",
        "/dev/nvme0n1p15       0      0       0     - /boot/efi",
        "",
        "Failed units:"
    ]
}
changed: [staging.righttoknow.org.au] => {
    "status_result.stdout_lines": [
        "ip-172-31-34-95",
        "Sun Aug 16 01:04:17 AEST 2026",
        "",
        " 01:04:17 up 4 days,  2:34,  1 user,  load average: 0.02, 0.08, 0.08",
        "",
        "               total        used        free      shared  buff/cache   available",
        "Mem:            3856        1432         381          12        2042        2105",
        "Swap:            511           4         507",
        "",
        "Filesystem      1M-blocks  Used Available Use% Mounted on",
        "/dev/root           19663 12412      7236  64% /",
        "/dev/nvme1n1         9980  7513      1940  80% /data",
        "/dev/nvme0n1p15       105     7        99   6% /boot/efi",
        "",
        "Filesystem       Inodes  IUsed   IFree IUse% Mounted on",
        "/dev/root       2580480 352600 2227880   14% /",
        "/dev/nvme1n1     655360 166376  488984   26% /data",
        "/dev/nvme0n1p15       0      0       0     - /boot/efi",
        "",
        "Failed units:",
        "* varnishncsa.service loaded failed failed Varnish HTTP accelerator log daemon"
    ]
}
ok: [web2.blue.planningalerts] => {
    "status_result.stdout_lines": [
        "ip-172-31-18-88",
        "Sun Aug 16 01:04:18 AEST 2026",
        "",
        " 01:04:18 up 4 days,  2:34,  1 user,  load average: 0.14, 0.20, 0.19",
        "",
        "               total        used        free      shared  buff/cache   available",
        "Mem:            3836        2404         196           2        1235        1157",
        "Swap:            511          46         465",
        "",
        "Filesystem      1M-blocks  Used Available Use% Mounted on",
        "/dev/root           15693  9049      6629  58% /",
        "efivarfs                1     1         1   4% /sys/firmware/efi/efivars",
        "/dev/nvme0n1p15       105     7        99   6% /boot/efi",
        "",
        "Filesystem       Inodes  IUsed   IFree IUse% Mounted on",
        "/dev/root       2064384 221956 1842428   11% /",
        "efivarfs              0      0       0     - /sys/firmware/efi/efivars",
        "/dev/nvme0n1p15       0      0       0     - /boot/efi",
        "",
        "Failed units:"
    ]
}
ok: [web.metabase.oaf.org.au] => {
    "status_result.stdout_lines": [
        "ip-172-31-38-4",
        "Sun Aug 16 01:04:18 AEST 2026",
        "",
        " 01:04:18 up 4 days,  2:34,  1 user,  load average: 0.09, 0.05, 0.01",
        "",
        "               total        used        free      shared  buff/cache   available",
        "Mem:            1930        1165          71           0         693         589",
        "Swap:            511         255         256",
        "",
        "Filesystem      1M-blocks  Used Available Use% Mounted on",
        "/dev/root            9741  9228       497  95% /",
        "/dev/nvme0n1p15       105     7        99   6% /boot/efi",
        "",
        "Filesystem       Inodes  IUsed   IFree IUse% Mounted on",
        "/dev/root       1290240 156274 1133966   13% /",
        "/dev/nvme0n1p15       0      0       0     - /boot/efi",
        "",
        "Failed units:"
    ]
}
ok: [au.proxy.oaf.org.au] => {
    "status_result.stdout_lines": [
        "ip-172-31-14-229",
        "Sun Aug 16 01:04:18 AEST 2026",
        "",
        " 01:04:18 up 4 days,  2:35,  1 user,  load average: 0.00, 0.00, 0.00",
        "",
        "              total        used        free      shared  buff/cache   available",
        "Mem:            466         149          12           5         304         279",
        "Swap:             0           0           0",
        "",
        "Filesystem     1M-blocks  Used Available Use% Mounted on",
        "/dev/nvme0n1p1      7877  2861      5001  37% /",
        "",
        "Filesystem      Inodes  IUsed  IFree IUse% Mounted on",
        "/dev/nvme0n1p1 1024000 141843 882157   14% /",
        "",
        "Failed units:"
    ]
}

PLAY RECAP ************************************************************************************************************************************************************************************************
52.64.33.12                : ok=0    changed=0    unreachable=1    failed=0    skipped=0    rescued=0    ignored=0   
au.proxy.oaf.org.au        : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
openaustralia.org.au       : ok=2    changed=2    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
prod.righttoknow.org.au    : ok=2    changed=2    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
srv.theyvoteforyou.org.au  : ok=2    changed=2    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
staging.righttoknow.org.au : ok=2    changed=2    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
web.metabase.oaf.org.au    : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
web1.blue.planningalerts   : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
web2.blue.planningalerts   : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

FAILED: Some servers failed the status check!

```

FYI - now uses `server-status.yml` playbook.

ianheggie-oaf:
> * `-l <group>` with an empty match is a fatal Ansible error, so a host disappearing entirely from the dynamic inventory fails loudly rather than silently no-opping. It's only the _partial_ case that's quiet (see the inline comment on `hostnames`)

As I rejected the suggestion to restrict inventory to only running instances, if the instance is present, but not responsive, then ansible will report it as an error but proceed with the instances that are responsive.

ianheggie-oaf:
Direct reply to @benrfairless 's comments:

> Review: Ansible over SSM for the EC2 fleet

> Blast radius: high. This changes how every EC2 host is discovered, 
- [x] Yes, deliberate "one place for everything and everything in its place" principle and aligning with intended dynamic discovery for capistrano for consistency

> renames CloudWatch log streams, and 
- [x] checked that this PR doesn't change the log names 

> moves the restic backup paths. 
- [ ]

> I traced the consumers rather than trusting the diff: there are no remaining inventory_hostname uses that carry identity, and the rds to rds_instance_info migration is complete across every consumer (group_vars/ec2.yml, site.yml, packer/playbook.yml). All 7 CI checks pass.
- [ ] 

> I've left inline suggestions for the fixable items. Leaving this as a comment rather than "request changes" - the one gating item is a verification run, not a code defect.
One thing I'd want confirmed before merge

> roles/internal/openaustralia/meta/main.yml:144 and roles/internal/righttoknow/meta/main.yml:154 now derive the restic repository path from log_name, and the test plan marks apply-* as not yet run for either service.

> I worked through the values and it should be a no-op: log_name equals the old inventory_hostname exactly for both (openaustralia.org.au and prod.righttoknow.org.au), which is what https://github.com/openaustralia/infrastructure/commit/7f311b53ca06ac3fb11814bd771c9f8f26d51e5a was fixing. But "should be" is the problem. If it isn't, restic initialises a fresh empty repository at a new S3 prefix and backup continuity for production data restarts from zero.

> Could you run make check-openaustralia and STAGE=production make check-righttoknow with --diff and confirm the restic profile and cron show no change to the repository path? Both are check-*, so read-only, no production go-ahead needed.
Not inline-able, so noting here

> AGENTS.md's known-safe read-only list covers show-* (so show-aws-inventory is fine) but not ssh-config or server-status. Per AGENTS.md's own rule, anything off that list gets treated as a live-fire production action needing an explicit go-ahead, which is a bit silly for two read-only helpers. This PR already edits AGENTS.md, so it's a small addition in a file it's touching anyway.
On the two open Sentry bot comments

> Both look wrong to me, flagging so they don't drive a change:

> bin/ssh-config:59, "User root fails, needs ubuntu" (HIGH). False positive. roles/internal/deploy-user authorises github_users' personal keys on root unrestricted; only the terraform key gets the command="Please login as the user ubuntu" forced command. ansible.cfg sets remote_user = root for the whole fleet. User root is correct. (One real caveat: a freshly built instance that hasn't been provisioned yet still has cloud-init's restriction, so root fails until the first Ansible run.)
> roles/internal/openaustralia/meta/main.yml:144, "should use public_hostname" (MEDIUM). The bot has the fix backwards. public_hostname is deliberately shared across a fleet, so using it for a restic repository would point two instances at one repo, i.e. concurrent writers on the same backup store. log_name is the right choice here. The underlying verification concern is real, and is the gating item above.

> Smaller things, all inline

> Alias collision in bin/ssh-config, 
> a silent-drop failure mode in the dynamic inventory, 
> a missing --profile on the inventory ProxyCommand, 
> bin/server-status aborting on a single unreachable host, 
> plus two comment/whitespace nits.

roles/internal/proxy/tasks/main.yml
```diff
@@ -3,11 +3,12 @@
   apt:
     pkg: build-essential
 
-- name: Download tinyproxy source code
+- name: "Download tinyproxy source code (change notification turned off)"
   get_url:
     url: https://github.com/tinyproxy/tinyproxy/releases/download/1.10.0/tinyproxy-1.10.0.tar.gz
     dest: /tmp/tinyproxy-1.10.0.tar.gz
   register: tinyproxy_source
```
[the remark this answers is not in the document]
ianheggie-oaf: Fixed the code

benrfairless (approved):
LGTM from a review and as discussed

PR #650, opened 2026-08-16
Remove OLD openaustralia server from terraform

ianheggie-oaf:
## Description

Removes `terraform/openaustralia/main.tf` entirely: the `aws_instance.main` ("OLD openaustralia"),
its `aws_eip.main`, and the `aws_ebs_volume.data`/`aws_volume_attachment.data` it used. This is a
separate, legacy resource block sitting alongside the real `openaustralia-prod` instance in
`terraform/openaustralia/production.tf` - it has no `AnsibleGroups` tag (so it was never part of
the Ansible-managed fleet), and its `key_name = "test"` and `Name = "OLD openaustralia"` both flag
it as defunct.

## Motivation and Context

Spotted while adding `Application`/`Roles` capistrano-aws tags to the real openaustralia instance
in `terraform/openaustralia/production.tf` (see #643) - this old resource has
no such tags and isn't part of the current fleet, so it's dead weight in the module rather than
something to tag.

Part of #678 (removing this instance alongside the unused openvpn server; the openvpn half is
tracked separately in #679).

## How Has This Been Tested?

- [x] `make tf-check-fmt` passes.
- [ ] Checked affected area manually on my own / staging system - not yet done, per the warning
      below: needs confirmation the instance/volume are truly unused before `tf-apply`
- [x] Ran automated tests on my own system (`make tf-check-fmt` and the rest of `make lint` pass)
- [ ] **`make tf-validate` currently fails on this branch** - see "Known issue found via testing"
      below before merging.

## Known issue found via testing

AWS credentials are available again in the environment this was tested in, so this is a real
result, not the earlier credential-expiry limitation noted in previous revisions of this
description. Deleting `terraform/openaustralia/main.tf` wholesale removed more than the "OLD
openaustralia" instance - that file also carried:

1. The `openaustralia` module's only `required_providers` block (for `cloudflare/cloudflare`).
   Every other module keeps this in its own `main.tf`. Without it, Terraform falls back to the
   legacy `hashicorp/cloudflare` registry namespace for this module, and `init` fails outright -
   which blocks `validate`/`plan`/`apply` for the **entire repo**, not just this module, so it
   breaks every PR later in this stack too (#680, #644, #645, #646).
2. `aws_ebs_volume.data`/`aws_eip.main`, which three places outside `main.tf` still referenced:
   - `production.tf`'s `aws_instance.production.availability_zone` pointed at
     `aws_ebs_volume.data.availability_zone` (the *old* instance's volume) instead of its own
     `aws_ebs_volume.production_data` - a latent copy-paste bug, currently silent on `main` only
     because both volumes happen to be hardcoded to the same AZ.
   - `outputs.tf` exported `main_public_ip`, wired through root `terraform/main.tf` into
     `module.oaf`'s `openaustralia_main_ip` variable, which is spliced into the **live SPF record**
     for `_spf1.oaf.org.au` (`terraform/oaf/dns.tf`) - shared by every OAF domain that includes it.
   - `openaustralia/dns.tf`'s `cloudflare_record.old_root` A record (`oldprod.openaustralia.org.au`)
     also pointed at the old instance's EIP - left in place, it would point at an Elastic IP AWS is
     free to reassign to someone else once released (a subdomain-takeover-shaped risk), not just a
     broken Terraform reference.

I've written and verified a fix for all four (`terraform validate` and `terraform fmt` both pass
with it applied, confirmed on this branch and re-confirmed with no conflicts at the tip of the
stack) but haven't pushed it here - two parts are content decisions, not just code fixes, so I'd
rather you confirm them than have me apply them silently:

- Dropping `oldprod.openaustralia.org.au` entirely, since its target no longer exists.
- Dropping the old server's IP from the shared SPF record, since it's decommissioned.

Both seem correct given #678's intent, but they touch live DNS/email deliverability for OAF's
domains, so flagging rather than deciding. Full diff available on request.

**Before merging/applying**: this instance and its EBS volume still exist in AWS and in Terraform
state. Applying this PR will actually terminate that EC2 instance, release its Elastic IP, and
delete the EBS volume with it - please confirm this is truly unused (nothing serving traffic from
it, no data on that volume worth keeping) before running `make tf-apply` on this.

N/A - Terraform-only change.

## Types of Changes

- [ ] Documentation

## Checklist:

---

**AI-assisted contribution**: this pull request was opened by Claude Code (model:
claude-sonnet-5) at the user's request, drafting this description from the existing commit
on the branch. The code change itself (the commit removing `terraform/openaustralia/main.tf`)
was authored directly by the human, not generated by AI.

---

Retargeted base from `main` to `feature/remove_openvpn` to reflect where this branch actually sits in the stack.


benrfairless (approved):
LGTM at a glance

ianheggie-oaf:
@benrfairless - please confirm if I need to backup or do anything else before destroying this instance and its associated EBS volumes:
- [ ] Yes - needs some backup and I'll say what and where to put it in a comment
- [ ] No - go ahead and apply the terraform changes which will destroy the instance, ebs volume and associated resources

Issue #653, opened 2026-08-17
Bug: righttoknow role has drifted from Alaveteli's config examples in 10 places


benrfairless:
### Which service is this about?

Right to Know

### What happened?

I compared every config example Alaveteli ships (`config/*example*`, plus `varnish-alaveteli.vcl`) against what `roles/internal/righttoknow` actually deploys, and against the theme's Capistrano config. Twelve items had drifted from upstream or had no counterpart at all.

Ten of them are collected here for triage: operational hygiene, upgrade blockers, and stale config. The numbering below is from the full audit and starts at 3 — items 1 and 2 are tracked separately in `openaustralia/oaf-internal` because they need handling before anything about them goes in a public repo.

**Baseline for the comparison.** Both the `staging` and `production` branches of `openaustralia/alaveteli` are at `0.45.8.0`, and no config example changed between the `0.45.8.0` tag and the current `staging` head. The local `release/0.46.7.0` branch is the pending upgrade, and it touches exactly four config examples.

**This is a repository-to-repository comparison.** I did not inspect the running servers, so everything below about current behaviour is inferred from configuration and wants confirming on the host before anyone acts on it. Items 3 and 7 in particular have a cheap on-host check noted against them.

---

## Process management: items 3, 4 and 6 are one piece of work

**3. `send-notifications` and `foi-alert-tracks` are installed but never enabled or started.** `tasks/main.yml:325` templates both into `/etc/init.d/` and stops there. There is no `service`/`systemd` task setting `enabled: true` or `state: started`, and no handler to restart them — unlike Sidekiq, which gets all three at `tasks/main.yml:228`.

They do come up on every deploy, because `deploy:restart` runs `/etc/init.d/alaveteli-<stage> restart` and that script restarts both. So the gap is reboot survival specifically: with no runlevel symlinks from `update-rc.d`, restarting the box without a deploy afterwards leaves email alerts and notifications stopped, silently.

Worth running `ls /etc/rc2.d/ | grep -E 'notif|alert'` on the hosts before assuming this is live — the symlinks may have been created by hand years ago, which would make it latent instead.

**4. Sidekiq is never restarted on deploy, so it keeps running the previous release's code.** The unit sets `WorkingDirectory=/srv/www/<stage>/current`, and the running process holds whatever code it loaded at boot. A deploy repoints `current` but nothing signals Sidekiq.

[...]

PR #657, opened 2026-08-17
Drop obsolete nginx ssl directive for RTK

benrfairless:
## Description

Two related nginx changes for Right to Know, covering the three HTTPS server blocks (`templates/nginx/default.j2`, `templates/nginx/stage.j2`) and the http-level defaults in `files/nginx/nginx.conf`:

- Removes the standalone `ssl on;` directive from all three server blocks. Each already carries `listen 443 ssl http2`, so it was redundant; nginx deprecated it in 1.15.0 and removed it in 1.25.1, where it becomes a config load failure. Clearing it now means a later distro nginx bump cannot take the site down on reload.
- Narrows `ssl_protocols TLSv1 TLSv1.1 TLSv1.2` to `ssl_protocols TLSv1.2 TLSv1.3` in the server blocks and the http-level default, so a future server block without its own override inherits the modern list. TLS 1.0/1.1 are deprecated by RFC 8996. The hosts run nginx 1.18 with OpenSSL 3.0 (confirmed `nginx/1.18.0` on both live hosts, read-only check), which accepts and can negotiate TLSv1.3.

Deliberately deferred: the `http2` flag on the `listen` lines stays. The replacement `http2 on;` directive only exists from nginx 1.25.1, so switching now would break the currently shipped nginx. That follows once the fleet crosses 1.25.1.

Context for reviewers: righttoknow.org.au and www are Cloudflare-proxied, so for public traffic `ssl_protocols` here governs the Cloudflare-to-origin hop; the unproxied prod./staging. hostnames are reached directly. Upstream's `nginx-ssl.conf.example` carries the same `ssl on;` line, so this is shared debt rather than something RTK introduced. The theyvoteforyou role has the same `ssl on;` in three templates, which deserves its own follow-up.

## Motivation and Context

Item 7 of #653: forward compatibility with nginx >= 1.25.1 plus TLS hardening. Part of openaustralia/infrastructure#653.

## How Has This Been Tested?

`make yaml-lint`, `make template-check` and `ansible-lint` (CI config) pass, byte-identical to a pre-edit baseline. nginx cannot be syntax-checked locally because the rendered config and certificates only exist on the hosts; an `nginx -t` on staging after `STAGE=staging make apply-righttoknow` is the real test.

## Types of Changes

- [ ] Documentation

## Checklist:

---

AI disclosure: this change was written by Claude Code (model `claude-opus-5`) as part of an orchestrated run coordinated by OpenCode (model `anthropic.claude-fable-5`), reviewed and committed by a human. See the `Assisted-by:` commit trailer.

### DIFF
--- a/roles/internal/righttoknow/files/nginx/nginx.conf
+++ b/roles/internal/righttoknow/files/nginx/nginx.conf
@@ -39,7 +39,7 @@ http {
 	# SSL Settings
 	##
 
-	ssl_protocols TLSv1 TLSv1.1 TLSv1.2; # Dropping SSLv3, ref: POODLE
+	ssl_protocols TLSv1.2 TLSv1.3;
 	ssl_prefer_server_ciphers on;
 
 	##
--- a/roles/internal/righttoknow/templates/nginx/default.j2
+++ b/roles/internal/righttoknow/templates/nginx/default.j2
@@ -15,11 +15,10 @@ server {
   server_name _;
   rewrite ^/(.*) https://www.{{ righttoknow_domain }}/$1 permanent;
 
-  ssl on;
   ssl_certificate /etc/letsencrypt/live/{{ righttoknow_domain }}/fullchain.pem;
   ssl_certificate_key /etc/letsencrypt/live/{{ righttoknow_domain }}/privkey.pem;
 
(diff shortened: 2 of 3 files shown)


ianheggie-oaf (approved):
Approved - lgtm at a quick read through

PR #658, opened 2026-08-17
Sync general.yml keys with upstream examples

benrfairless:
## Description

Brings `roles/internal/righttoknow/templates/general.yml.j2` back in step with upstream Alaveteli's `general.yml-example`, comment blocks included:

- Adds `BLOCK_SPAM_USER_MESSAGES: true`. **This is a behaviour change and the one thing in this PR that needs a deliberate call**: upstream defaults it to false, the key already exists in the deployed 0.45.8, and every other `BLOCK_SPAM_*` toggle in this template is explicitly true, so it was set true to match. Flip to `false` before merging if the current behaviour should be kept instead.
- Adds three settings that arrive with 0.46.7.0, each set to the upstream default so behaviour is unchanged: `READ_ONLY_FEATURES: []` (granular read-only control, useful during a spam wave), `ENABLE_PUBLIC_ANNOTATIONS: true` and `ENABLE_USER_TO_USER_MESSAGING: true` (both moderation levers worth pinning explicitly rather than trusting a default upstream could change). Alaveteli ignores unknown `general.yml` keys, so these are safe to deploy ahead of the upgrade.
- Removes the dead `BACKGROUND_JOBS: 'server'` key and its comment block. Alaveteli removed the setting; nothing reads it.

Placement note: `BLOCK_SPAM_USER_MESSAGES` sits after `BLOCK_SPAM_COMMENTS`, matching where upstream puts it in both 0.45.8 and 0.46.7.

## Motivation and Context

Items 8, 9 and part of 12 of #653. Of 112 top-level keys in upstream's 0.45.8 example, `BLOCK_SPAM_USER_MESSAGES` was one of only two absent from this template with no recorded reason. Part of openaustralia/infrastructure#653.

## How Has This Been Tested?

`make yaml-lint`, `make template-check` and `ansible-lint` (CI config) pass. The template was additionally parsed as YAML with the Jinja expressions neutralised: 114 keys, all four new values present and correct, `BACKGROUND_JOBS` absent.

## Types of Changes

- [ ] Documentation

(`BLOCK_SPAM_USER_MESSAGES: true` changes runtime behaviour for user-to-user messages that look like spam; everything else is behaviour-neutral.)

## Checklist:

---

AI disclosure: this change was written by Claude Code (model `claude-opus-5`) as part of an orchestrated run coordinated by OpenCode (model `anthropic.claude-fable-5`), reviewed and committed by a human. See the `Assisted-by:` commit trailer.

### DIFF
--- a/roles/internal/righttoknow/templates/general.yml.j2
+++ b/roles/internal/righttoknow/templates/general.yml.j2
@@ -466,6 +466,29 @@ SECRET_KEY_BASE: '{{ alaveteli_secret_key_base }}'
 # ---
 READ_ONLY: ''
 
+# READ_ONLY_FEATURES - Array of strings (default: [])
+#
+# Granular read-only control for specific features. This allows you to disable
+# specific functionality without putting the entire site in read-only mode.
+#
+# Available features:
+#   'annotations'     - Blocks annotations on requests
+#   'classifications' - Blocks request status classifications
+#   'followups'       - Blocks followup messages on requests
+#   'requests'        - Blocks new FOI requests
+#   'signups'         - Blocks new user registration
+#
+# Examples:
+#
+# Block only new requests:
+#   READ_ONLY_FEATURES: ['requests']
+#
+# Block new requests and annotations:
+#   READ_ONLY_FEATURES: ['requests', 'annotations']
+#
+# ---
+READ_ONLY_FEATURES: []
+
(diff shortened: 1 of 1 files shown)


ianheggie-oaf (approved):
Approved - lgtm based on reasoning given

PR #660, opened 2026-08-18
Stop RTK sysvinit daemons without their init script

benrfairless:
## Description

`make apply-righttoknow STAGE=all` fails on both hosts at the "Stop the sysvinit daemons replaced by systemd units" task, which was added in #656. The old daemons are left running and everything after that task, including the systemd unit installs and the database tasks, never runs.

The init scripts that task calls are the ones this repo used to install, and their stop function is `start-stop-daemon --stop --oknodo --retry 5 --pidfile "$PIDFILE"` with no second match criterion. Since dpkg 1.19.3, start-stop-daemon refuses that when the pidfile was written by a non-root user, and these daemons run as `deploy`, so the call aborts with `matching only on non-root pidfile /srv/www/<stage>/current/tmp/pids/<daemon>.pid is insecure` and the script exits 1.

That also explains why each host only failed on its own stage: the other stage's pidfile doesn't exist, so there was nothing to stat and `--oknodo` returned 0.

This calls `start-stop-daemon` directly instead of going through the broken script, adding `--user deploy` as the second match criterion. That is the remedy the Debian man page points to, and the same fix Alaveteli made to its own daemon templates upstream in November 2021 (alaveteli ac54334) - the scripts on our servers predate it.

The `removes:` guard is unchanged, so this stays a one-shot that no-ops once the following task has removed the init scripts. Nothing else invokes them: there are no cron entries referring to them in this repo, in Alaveteli's `config/crontab-example`, or in the theme.

One new failure mode worth knowing when reading a red run: with `--retry 5`, start-stop-daemon exits 2 if a process survives TERM, 5s, KILL, 5s. That means the daemon would not die, which is a different problem from this one.

## Motivation and Context

Provisioning Right to Know is currently blocked - the play aborts partway through on both staging and production.

Part of #653, and fixes the task added in #656.

## How Has This Been Tested?

`make yaml-lint` and `make ansible-lint` both pass locally.

Not yet applied to a host. `make check-righttoknow` can't verify this one, because `command` tasks are skipped in check mode. Before merging I'll confirm on staging that the pidfiles point at live daemons rather than recycled pids, using the no-op dry run:

```
/sbin/start-stop-daemon --stop --test --oknodo \
  --pidfile /srv/www/staging/current/tmp/pids/send-notifications.pid --user deploy
ps -p <the pid it names> -o pid,user,cmd=
```

then run `STAGE=staging make apply-righttoknow` before production, since everything after the failing task has never run on either host.

## Types of Changes

- [ ] Documentation

## Checklist:

---

Assisted by Claude Code using the claude-opus-5 model: it diagnosed the failure, traced it to the deleted init script templates and the upstream Alaveteli fix, and drafted the change and this description. Reviewed before submitting.


ianheggie-oaf (approved):
Approved - LGTM reading desc and diff

PR #663, opened 2026-08-19
Revert "Report RTK cron jobs and config to Sentry"

benrfairless:
Reverts openaustralia/infrastructure#661

This will be re-introduced when we are ready to deploy it.


ianheggie-oaf (approved):
Approved - LGTM, redo when u r ready

PR #664, opened 2026-08-19
Report RTK cron jobs and config to Sentry

benrfairless:
Re-lands #661, which was reverted in #663 until we were ready to deploy it. The change is identical (a cherry-pick of the original commit); the PR body below is duplicated from #661.

---

## Description

Provisioning half of adding Sentry to Right To Know (openaustralia/righttoknow#1004). The application code lives in the theme repo (openaustralia/righttoknow#1058); this supplies the config it reads and wraps the scheduled jobs:

- Renders `SENTRY_DSN` into each stage's `general.yml` from the existing vaulted `sentry_dsn` var, which was added for every service in 4da6654a but nothing consumed until now. Staging and production already hold separate keys for the same project, so the stages stay apart. Also renders `SENTRY_TRACES_SAMPLE_RATE` and `SENTRY_JS_LOADER_URL` (both new role defaults), so sampling and browser monitoring can be changed with a provision run rather than a deploy.
- Wraps all 30 RTK cron jobs in `sentry monitor run`, which reports a check-in when a job starts and finishes. That covers the failure exception email cannot: a job that stops running altogether. Each job passes its own schedule, so the monitors are created and kept in step from `cron.yml` rather than configured by hand in Sentry. Check-in margins and max runtimes are set per cadence; the two sub-hourly jobs need two consecutive failures before an issue is raised.
- Adds `rtk-cron-lock`, a wrapper that maps `run-with-lockfile -n`'s exit 100 (previous run still holds the lock) to 0. That skip is routine, not a failure - update-xapian-index runs every five minutes and will overlap - and without it Sentry would record every skip as a failed check-in. All other exit codes pass through, so real failures still register.
- Installs the Sentry CLI v4 (`sentry`, the rewritten CLI, not the older `sentry-cli`) as a pinned, checksummed binary from GitHub releases rather than piping the install script to bash. The version and sha256 are role defaults; note the tool's pre-1.0 versioning means breaking changes can land in a minor release, so bumps should be deliberate.

## Motivation and Context

We constantly hit monthly limits with Honeybadger and it no longer offers an OSS plan; Sentry has given OAF a sponsored plan. See openaustralia/righttoknow#1004. This also gives RTK its first real monitoring of the cron jobs, whose only current failure signal is the MAILTO email (and only when a job produces output).

## How Has This Been Tested?

- `make yaml-lint`, `make template-check` and `make ansible-lint` all pass locally (re-run against current `main` after the cherry-pick).
- The `general.yml.j2` changes were rendered through Jinja with stand-in values and parsed as YAML to confirm the new keys come out with the right types (string DSN and loader URL, float sample rate).
- The exact `sentry monitor run` invocation was exercised for real against the right-to-know Sentry project before the cron entries were written: a check-in upserted a monitor with the schedule, margin, max runtime and environment, and reported ok. (That left a throwaway `rtk-sentry-cli-smoke-test` monitor which needs deleting in the Sentry UI.)
- `run-with-lockfile.sh` was read to confirm the exit-100-on-held-lock behaviour that `rtk-cron-lock` translates.
- Not yet run against a server: the plan is to provision staging first, confirm the rendered `general.yml` (in particular that the vaulted `sentry_dsn` decrypts to the right-to-know project - I could not read the vault to check) and watch check-ins arrive, before provisioning production.

## Types of Changes

- [ ] Documentation

## Checklist:

---

AI disclosure: this change was developed with opencode using the model amazon-bedrock/anthropic.claude-fable-5, and re-landed (cherry-pick, lint re-run and this PR) the same way. The `Assisted-by` commit trailer names an incorrect model (anthropic/claude-opus-5) due to a drafting error; this note is the accurate record. A human reviewed, committed and signed off the work.

---

## Related pull requests

This is one release in three parts. All three need to merge and go out together, in this order:

1. openaustralia/infrastructure#675 - sets `ruby_version_production` to 3.4.10. Merge, then provision production so Ruby 3.4.10 is installed and `shared/rbenv-version` is updated.
2. openaustralia/infrastructure#664 - renders `SENTRY_DSN`, `SENTRY_TRACES_SAMPLE_RATE` and `SENTRY_JS_LOADER_URL` into production `general.yml`, and wraps the RTK cron jobs in `sentry monitor run`. Merge, then provision production.
3. openaustralia/righttoknow#1059 - the theme release, `staging` into `production`, deployed with Capistrano once both provision runs are done.

Why the order matters: production currently runs Ruby 3.2.9 and the release branch is on 3.4.10, so deploying #1059 before #675 has been provisioned would bundle against the wrong Ruby. Sentry is less severe, deploying #1059 before #664 is safe but inert, since `lib/sentry_config.rb` leaves the SDK disabled when no DSN is configured, so errors, traces and cron check-ins would silently go nowhere until production is provisioned.


ianheggie-oaf (approved):
Approved - LGTM with a quick read over

Issue #203, opened 2025-08-27
[Feature] Update nginx config to allow maintenance page to be configured


benrfairless:
### Background
Alaveteli has the below tasks available that allow us to display a maintenance message:
```ruby
cap deploy:web:disable               # Present a maintenance page to visitors.
cap deploy:web:enable                # Makes the application web-accessible again.
```

In order for this to work, an update is required to nginx.conf:

```nginx
 # if you are using Nginx add this to your server config:

        if (-f $document_root/system/maintenance.html) {
          return 503;
        }
        error_page 503 @maintenance;
        location @maintenance {
          rewrite  ^(.*)$  /system/maintenance.html break;
          break;
        }
```

PR #682, opened 2026-08-19
Serve RTK maintenance page nginx never showed

### LINKED ISSUE #681: Bug: nginx never serves Right to Know's maintenance page during migrations
### Which service is this about?

Right to Know

### What happened?

`roles/internal/righttoknow/templates/nginx/stage.j2:60` tests for `$document_root/system/maintenance.html`, which resolves to `/srv/www/<stage>/current/public/system/maintenance.html`. Nothing writes a file at that path, and no `system/` directory exists in Alaveteli's `public/`, in the theme's `public/`, or in the shared directories the role creates.

The deploy, which lives in `openaustralia/righttoknow`, uses capistrano-maintenance and deliberately writes the page somewhere else (`config/deploy.rb:28-29`, added in 2a3ac4a):

```ruby
set :maintenance_basename, 'down'
set :maintenance_dirname,  -> { current_path.join('public') }
```

So `maintenance:enable` writes `/srv/www/<stage>/current/public/down.html`. That follows the `down.html` convention Alaveteli itself uses: upstream's `config/nginx.conf.example:25` has `try_files /down.html $uri/index.html $uri @alaveteli;`, `config/httpd.conf-example` has the Apache equivalent, and `/public/down.html` is in Alaveteli's `.gitignore`. Our vhost has no `try_files` line and no `@alaveteli` location, because it runs Passenger straight off `root`, so nothing in nginx looks at `down.html` either.

The result is that `before 'deploy:migrate', 'maintenance:enable'` (`deploy.rb:209`) writes the page on every migrating deploy, nginx ignores it, visitors keep hitting the app while migrations run, and `maintenance:disable` tidies it away afterwards. The righttoknow README documents `cap <stage> deploy:migrate` as "maintenance page shown automatically", which is not currently true, and the same applies to running `maintenance:enable` by hand before something long like a Xapian rebuild.

The `system/maintenance.html` path arrived with #203 (972be095), when the deploy was still Capistrano 2 in the Alaveteli fork and `deploy:web:disable` wrote to `shared/system/maintenance.html`. It would not have lined up then either: Capistrano 2 only bridges the two by symlinking `current/public/system` to `shared/system` through its default `shared_children`, and that fork's `config/deploy.rb` replaced the list with the basenames of `SHARED_DIRECTORIES`, which has no `public/system` entry. So this has most likely never worked, which is presumably why nobody noticed the path was wrong.

### Steps to reproduce

1. From an `openaustralia/righttoknow` checkout, run `bundle exec cap staging maintenance:enable REASON="testing"`.
2. Confirm the file landed: `ls -l /srv/www/staging/current/public/down.html` on the staging host.
3. Load `https://www.<staging domain>/` (staging basic auth credentials needed). The application serves normally instead of a maintenance page.
4. `bundle exec cap staging maintenance:disable`.

### What did you expect to happen?

While `down.html` exists, requests get a 503 carrying the maintenance page. The smallest change in `stage.j2` is to keep the existing 503 mechanism and point it at the file the deploy actually writes:

```nginx
# Serve the maintenance page capistrano-maintenance writes (see
# config/deploy.rb in openaustralia/righttoknow).
location / {
  if (-f $document_root/down.html) {
    return 503;
  }
}

error_page 503 @maintenance;

location @maintenance {
  add_header Cache-Control "no-cache, no-store, must-revalidate" always;
  rewrite ^(.*)$ /down.html break;
}
```

Two things belong in the same change rather than as follow-ups:

**Keep the ACME challenge path reachable.** The test currently sits at server level, where nginx evaluates it in the rewrite phase before choosing a location, so with maintenance on `/.well-known/acme-challenge/` would get the maintenance page too and any renewal landing in the window would fail validation. A migration window is short enough that this is unlikely; a manual `maintenance:enable` before a Xapian rebuild is not. Moving the test into `location /` leaves the `^~ /.well-known/acme-challenge/`, `^~ /assets/` and `= /favicon.ico` locations alone. Restructuring locations in a Passenger vhost warrants an `nginx -t` and a staging check rather than trusting the snippet above.

**Don't let Varnish hold on to the 503.** The HTTPS `www` block proxies to Varnish, which comes back to nginx on 8000. `files/default.vcl`'s `vcl_backend_response` sets no TTL for a 503, so it falls through to the builtin VCL and takes `default_ttl`, and `files/varnish.service` starts `varnishd` without `-t`, leaving that at the 120s default. A cached maintenance response could then outlive `maintenance:disable` by up to two minutes. An explicit `Cache-Control: no-store` should make Varnish treat it as hit-for-pass instead. Worth confirming with `varnishlog` on staging rather than taking the fall-through on trust.

Note also that pages already in the Varnish cache keep being served to anonymous visitors during maintenance, so maintenance mode is not absolute. That is probably fine for a short migration window. Making it absolute is a VCL change and a separate decision.

### Anything else?

- Keeping the 503 rather than adopting Alaveteli's `try_files /down.html ...` is deliberate: `try_files` serves the down page with a 200, and our vhost has no `@alaveteli` upstream location to fall through to. Closing that gap properly belongs with the rest of the drift in #653.
- `stage.j2` is shared by staging and production, so the change lands on both.
- Out of scope here, but worth a follow-up in `openaustralia/righttoknow`: `maintenance_template_path` is not set, so the page rendered is capistrano-maintenance's own generic template rather than Alaveteli's `public/down.default.html` or anything RTK-branded. That only becomes visible once nginx serves the file at all.
- Related: #203 (added the rule being fixed), #653 (drift from Alaveteli's config examples).
- Checked against capistrano-maintenance 1.2.1 (MIT), `lib/capistrano/tasks/maintenance.rake`, and `openaustralia/righttoknow` at ad3df3e on `staging`. I did not inspect the running servers, so the behaviour described is inferred from configuration and wants confirming on the host.

Drafted with AI assistance (Claude Code: claude-opus-5).

benrfairless:
## Description

Fixes the Right to Know nginx vhost so the maintenance page is actually served during migrating deploys. Three changes to `roles/internal/righttoknow/templates/nginx/stage.j2`, all in the `www` block on port 8000:

- Point the maintenance test at `$document_root/down.html`, the file capistrano-maintenance in `openaustralia/righttoknow` actually writes (`maintenance_basename 'down'`, `maintenance_dirname current/public`), instead of `system/maintenance.html`, which nothing ever writes.
- Move the test from server level into `location /`, so `/.well-known/acme-challenge/`, `/assets/` and `/favicon.ico` stay reachable while maintenance is on — a cert renewal landing in a maintenance window would otherwise fail validation.
- Add `Cache-Control: no-cache, no-store, must-revalidate` to the 503 so Varnish treats it as hit-for-pass rather than caching it for `default_ttl` (120s), which could keep the maintenance page up for up to two minutes after `maintenance:disable`.

`stage.j2` is shared by staging and production, so the change lands on both at the next `apply-righttoknow`.

## Motivation and Context

Resolves #681. The check has pointed at the wrong path since #203 and has most likely never worked, so `before 'deploy:migrate', 'maintenance:enable'` writes a page nginx ignores and visitors keep hitting the app while migrations run.

## How Has This Been Tested?

`make template-check` passes. Rendered the template locally (both variants: with/without `certbot_webroot` and `password_protect`) and verified with `nginx -t` on nginx 1.31.3 — syntax ok; only warnings are the pre-existing `listen ... http2` deprecations. Also ran a live local nginx against the rendered config with a stub docroot (Passenger directives stubbed, not available in Homebrew nginx): with `down.html` present, `/` and deep paths return 503 with the page body and `Cache-Control: no-cache, no-store, must-revalidate`; `/.well-known/acme-challenge/`, `/favicon.ico` and `/assets/` still return 200; removing `down.html` ends the 503s. Still wants a staging check via `STAGE=staging make apply-righttoknow` then `cap staging maintenance:enable` / `maintenance:disable` before production, since local testing cannot cover Passenger or Varnish interaction.

## Types of Changes

- [ ] Documentation

## Checklist:

---

Drafted with AI assistance (Claude Code: anthropic.claude-fable-5).


benrfairless:
## Local test evidence

Tested the rendered template on nginx 1.31.3 (Homebrew, macOS). Method:

1. Rendered `stage.j2` with Jinja2 using representative values (`domain=righttoknow.org.au`, `stage=staging`), in two variants: with `certbot_webroot` + `password_protect=true`, and with neither.
2. Substituted local-only stubs: `passenger_*` directives commented out (no Passenger module in Homebrew nginx) and the Let's Encrypt cert paths swapped for a throwaway self-signed cert. Neither touches the maintenance logic under test.
3. Ran `nginx -t` on both variants: **syntax ok**. Only warnings are the pre-existing `listen ... http2` deprecations (nginx ≥ 1.25 deprecates that form; not part of this change).
4. Started nginx on local ports with a stub docroot and exercised the port-8000 `www` server block over HTTP.

Results with `down.html` present (maintenance on):

| Request | Expected | Got |
|---|---|---|
| `/` | 503 + page body + no-store | **503**, body served, `Cache-Control: no-cache, no-store, must-revalidate` |
| `/request/some_request` | 503 | **503** |
| `/.well-known/acme-challenge/tok` | 200, token body | **200**, token served |
| `/favicon.ico` | 200 | **200** |
| `/assets/favicon.ico` | 200 | **200** |

After removing `down.html`: `/` no longer returns 503 (403 locally, but only because there's no Passenger to hand the request to — on the real host it falls through to the app).

That confirms all three behaviours of the fix: the corrected filename, the exempted ACME/assets/favicon locations, and the anti-caching header on the 503.

Not covered locally, still wanted on staging before production: Passenger interaction in the new `location /` (inheritance of `passenger_enabled` from server level), and Varnish treating the 503 as hit-for-pass (`varnishlog` check per the issue).

Testing drafted with AI assistance (Claude Code: anthropic.claude-fable-5).

ianheggie-oaf (approved):
Approved - LGTM - Good to know they have changed the path for maint page that had been the same for a decade or so

Issue #445, opened 2026-04-22
Fix theyvoteforyou ansible playbook (later)


ianheggie-oaf:
**Describe the bug**

Ansible provisioning fails at: Apt add passenger to list - signatures couldn't be verified because the public key is not available

**To Reproduce**

In infrastructure project:
```
make vagrant
vagrant up --provision theyvoteforyou.test

fails: ok=42 FAILED=1 - external/rvm.ruby : Update rvm (will retry)
```

**Expected behavior**

failed=0 at summary at end

**Screenshots/Output**

**Additional context**

ianheggie-oaf:
@Br3nda and @benrfairless - made into an independent issue

Issue #96, opened 2018-09-28
Can't use development without the Ansible Vault password


benrfairless:
As a very junior developer that is eager to start working on upgrading Right to Know soon, it's proving difficult as I need the Ansible Vault password. 

If I don't have it, then it's not possible to spin up the development environment.

It should be possible for volunteers and contributors to work on these projects without needing to rip out all the encrypted bits.

PR #684, opened 2026-08-20
Proposal for discussion: dev hosts for testing Ansible without Vagrant

ianheggie-oaf:
## Description

Adds one document, `docs/dev-hosts-implementation-plan.md`. No code, and nothing in `docs/adr/` is touched.

It is **a proposal for discussion, not an agreed plan**, in two halves:

1. Implementation detail: the work list, the compose mechanics, and what still needs checking.
2. After the horizontal rule, a proposed replacement body for `docs/adr/0001`, in a fenced block so it can be
   argued over a line at a time before the real ADR is edited.

The headline: Docker/Compose development hosts for most services, plus a rented-cloud backend for the cases
containers cannot serve. Three independent reasons the second backend is needed, any one sufficient: some
services have no arm64 path at all, Docker-in-Docker cannot be combined with architecture emulation, and
containers cannot be rebooted, which rules out testing whether a service starts on boot.

Two decisions are deliberately left open for this discussion: which cloud provider, and what to do about
openaustralia's pinned Xapian version.

## Motivation and Context

Closes nothing yet; proposes how to address #555, where Vagrant + VirtualBox does not work on Apple Silicon and
so blocks a contributor entirely. Also bears on #96 (vault passphrases block contributors from any local
testing) and #445 (the theyvoteforyou playbook is currently broken, which matters for telling new breakage from
old).

ADR 0001 already decided to leave Vagrant for Docker/Compose, but assumed containers could serve every service.
Investigating that found they cannot, which is why 0001 needs updating rather than just implementing.

## How Has This Been Tested?

Documentation only, so there is nothing to run. `Vagrantfile` is untouched and still works on Linux x86_64; it
stays until a service is proven on the new path.

Nothing here has been run on Apple Silicon. The document ends its first half with a prioritised list of five
checks that need a Mac.

- [ ] Ran automated tests on my own system

## Types of Changes

- [x] Documentation

## Checklist:

- [x] My change requires a change to the documentation.

## Sign-off needed

Two separate approvals before any of this is implemented: the technical direction, and recurring third-party
spend for the cloud backend.

---

Assisted by Claude Code (claude-opus-5). Architecture and pricing claims were checked against vendor sources;
anything unconfirmed is marked as such in the document.

### DIFF
--- a/docs/dev-hosts-implementation-plan.md
+++ b/docs/dev-hosts-implementation-plan.md
@@ -1,0 +1,535 @@
+# Development hosts: testing Ansible on Linux and macOS
+
+**This is a proposal for discussion, not an agreed plan.** Suggested edits on the draft pull request are the point
+of it.
+
+**Goal:** let both Linux on `x86_64` and macOS on `arm64` test the Ansible playbooks in this repository.
+
+**Status:** nothing started. `Vagrantfile` still works on Linux `x86_64` and stays until a service is proven on the
+new path, so nobody loses a working setup while this is discussed.
+
+This document has two halves. The first is the proposal, structured as four decisions that can be taken
+independently. The second, after the horizontal rule near the end, is a **proposed replacement for
+[`docs/adr/0001`](adr/0001-local-ansible-testing-moves-to-docker.md)**, sitting here rather than in `docs/adr/` so
+it can be argued over a line at a time before the real ADR is touched.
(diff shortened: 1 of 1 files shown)


ianheggie-oaf:
@benrfairless - for discussion as it mainly impacts you, but does me for full morph dev

PR #712, opened 2026-08-28
Fix over-deep staging www hostname for Cloudflare wildcard cert

### LINKED ISSUE #711: Bug: cannot turn proxy on for www.staging.rightoknow.org.au as standard wildcard ssl doesnt match on cloudflare
### Which service is this about?

Right to Know

### What happened?

When I turned proxying on for www.staging.righttoknow.org.au to test if remote_ip is correct chrome and firefox rejected it because cloudflare doesn't provide a wildcard match for *.*.DOMAIN, only *.DOMAIN 

### Steps to reproduce

1. Browse to https://www.staging.righttoknow.org.au/
2. Notice the error

### What did you expect to happen?

To be able to access website contents without error

### Browser, device, or OS (if relevant)

Firefox 154.0.1 (64-bit) and Google Chrome Version 151.0.7922.173 (Official Build) (x86_64) on Ubuntu 24.04.4 LTS

### Screenshots or error messages

Error on firefox:
```
Be careful. Something doesn’t look right.

Firefox can’t create a secure connection to the server at www.staging.righttoknow.org.au.
Advanced
What makes the site look dangerous?

It looks like this site is using old software with known security issues.
What can you do about it?

Make sure you’re using the latest version of Firefox. Go to Help > About Firefox in the menu. If you’re using the latest Firefox, the problem is most likely with the site itself.

[Learn more about secure connection failures](https://support.mozilla.org/1/firefox/154.0.1/Linux/en-GB/connection-not-secure)

Error Code: SSL_ERROR_NO_CYPHER_OVERLAP
```

### Anything else?

**I have applied the change in:**
* https://github.com/openaustralia/infrastructure/pull/712

So www.staging no longer exists, it is replaced with www-staging.righttoknow.org.au
```
ianh@ben:.../infrastructure (feature/711-rtk-stagi)$ host www-staging.righttoknow.org.au
www-staging.righttoknow.org.au is an alias for staging.righttoknow.org.au.
staging.righttoknow.org.au has address 13.237.220.112
staging.righttoknow.org.au mail is handled by 1 smtp.google.com.
ianh@ben:.../infrastructure (feature/711-rtk-stagi)$ host www.staging.righttoknow.org.au
Host www.staging.righttoknow.org.au not found: 3(NXDOMAIN)
```

ianheggie-oaf:
## Description

Adds a `www_prefix` Ansible variable for the Right to Know role/terraform module: `"www."` for production, `"www-"` for staging. Every place that built a `www` hostname by hardcoding `www.` now uses `{{ www_prefix }}{{ domain }}` (or `righttoknow_domain`) instead. Because each prefix carries its own separator, this reproduces today's production hostname (`www.righttoknow.org.au`) exactly, while giving staging a single-label hostname (`www-staging.righttoknow.org.au`) instead of the current two-label one.

Also renames the corresponding Cloudflare CNAME record in `terraform/righttoknow/dns.tf` from `www.staging.righttoknow.org.au` to `www-staging.righttoknow.org.au` (same resource address, only the `name` attribute changes, confirmed via `tf-check-fmt`/`tf-validate` that this doesn't force a replace).

Turns proxying fully one for staging rather than manually on for staging but not www.staging (now www-staging) in cloudflare and not on in `terraform/rightoknow./dns.tf`.

Documents PROVIDE_WHATISMYIP in general.yml.j2 template related to:
* openaustralia/righttoknow#1076

## Motivation and Context

Fixes #711. `www.staging.righttoknow.org.au` is two subdomain labels below the registered domain (`righttoknow.org.au`), which falls outside Cloudflare's Universal SSL wildcard coverage (apex + one level: `righttoknow.org.au` / `*.righttoknow.org.au`). This currently only "works" because staging's `www` record is grey-clouded and certs are issued via Let's Encrypt DNS-01, not Cloudflare's edge cert - but it would block staging from ever being moved behind the Cloudflare proxy.
* #711

Companion PR: openaustralia/righttoknow#1080 corrects the staging URL documented in that repo's README to match the new `www-staging.righttoknow.org.au` hostname.
* openaustralia/righttoknow#1080

## How Has This Been Tested?

- [x] Ran automated tests on my own system (`make ansible-lint`, `make yaml-lint`, `make template-check`, `make tf-check-fmt`, `make tf-validate` all pass locally)

**This HAS been applied** to staging!
 - `make tf-plan-target MODULE=righttoknow` to confirm only the DNS record's `name` changes in place, then a coordinated 
 - Applied with `tf-plan-target MODULE=righttoknow`
 - `STAGE=staging make apply-righttoknow` so the new hostname's cert (SAN list) and nginx config where updated

## Types of Changes

- [ ] Documentation

## Checklist:

---

Assisted by Claude Code (claude-sonnet-5).
Directed by and checked by @ianheggie-oaf

### DIFF
--- a/group_vars/righttoknow_staging.yml
+++ b/group_vars/righttoknow_staging.yml
@@ -6,6 +6,10 @@ site_name: staging.righttoknow.org.au
 righttoknow_domain: staging.righttoknow.org.au
 domain: staging.righttoknow.org.au
 server_hostname: staging.righttoknow.org.au
+# A single label ("www-staging", not "www.staging") so the hostname stays within
+# Cloudflare's Universal SSL wildcard coverage (apex + one level) if this is ever
+# proxied - see issue #711.
+www_prefix: "www-"
 
 # Cron enable/disable switch for staging
 cron_enabled: true
--- a/group_vars/righttoknow_production.yml
+++ b/group_vars/righttoknow_production.yml
@@ -7,6 +7,7 @@ site_name: prod.righttoknow.org.au
 # Domain configuration for standalone production server
 righttoknow_domain: righttoknow.org.au # Used in certbot and SSL
 domain: righttoknow.org.au # Used in Nginx
+www_prefix: "www." # Prepended to domain/righttoknow_domain to build the www hostname
 server_hostname: prod.righttoknow.org.au # Used in Nginx and certbot
 
 # First sentry key (for production only)
--- a/roles/internal/righttoknow/tasks/certificates.yml
+++ b/roles/internal/righttoknow/tasks/certificates.yml
@@ -43,18 +43,18 @@
   set_fact:
     staging_certs:
       - email: 'contact@oaf.{{ base_domain }}'
-        domains: 
+        domains:
           - "{{ righttoknow_domain }}"
-          - "www.{{ righttoknow_domain }}"
+          - "{{ www_prefix }}{{ righttoknow_domain }}"
   when: "'righttoknow_staging' in group_names"
 
 - name: Set certificate configuration for production
   set_fact:
     production_certs:
       - email: 'contact@oaf.{{ base_domain }}'
-        domains: 
+        domains:
           - "{{ righttoknow_domain }}"
-          - "www.{{ righttoknow_domain }}"
+          - "{{ www_prefix }}{{ righttoknow_domain }}"
           - "{{ server_hostname }}"
   when: "'righttoknow_production' in group_names"
 
--- a/roles/internal/righttoknow/templates/general.yml.j2
+++ b/roles/internal/righttoknow/templates/general.yml.j2
@@ -1387,3 +1387,8 @@ SENTRY_TRACES_SAMPLE_RATE: {{ sentry_traces_sample_rate }}
 #
 # ---
 SENTRY_JS_LOADER_URL: '{{ sentry_js_loader_url }}'
+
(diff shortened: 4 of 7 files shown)


group_vars/righttoknow_staging.yml
```diff
@@ -6,6 +6,10 @@ site_name: staging.righttoknow.org.au
 righttoknow_domain: staging.righttoknow.org.au
 domain: staging.righttoknow.org.au
 server_hostname: staging.righttoknow.org.au
+# A single label ("www-staging", not "www.staging") so the hostname stays within
+# Cloudflare's Universal SSL wildcard coverage (apex + one level) if this is ever
+# proxied - see issue #711.
+www_prefix: "www-"
```
Br3nda, line 12: did you look into if we can have staging.righttoknow.org.au directly? (and not need the `www-` version) ? That seems like it would get us to the necesary level to make Cloudflare happy.

Br3nda: actually, your way of having a prefix on both is better.. even if it looks a bit cumbersome

Br3nda (approved):

group_vars/righttoknow_staging.yml
ianheggie-oaf: It was the best i could think of, since I didn’t want some of the config to be only present for production.

PR #713, opened 2026-08-29
deploy-user: fix corrupted authorized_keys (blank line + YAML-folding bug)

Br3nda:
## What

`Set authorized_keys for local accounts` was failing with `invalid key specified` for the `deploy` and `ubuntu` accounts, and silently corrupting `root`'s authorized_keys (two keys glued onto one line, with a literal `\n` between them instead of a real line break).

## Why - two separate bugs stacked here

1. **`github_keys`'s trailing newline.** Its for-loop template's last content line (`{{ line }} {{ result.item }}`) ends with a real newline of its own - `trim_blocks` only trims the newline right after a `{% %}` tag, not this one. Joining that with the rest via `join('\n')` left a blank line right after the GitHub users' keys, which the `authorized_key` module rejects - and then reports every key after that point as part of the one bad entry, which is why the error bundled the CI deploy keys and the terraform key together for `deploy`.

2. **The `>-` folded block scalar.** `permitted_keys_for_account` was written across several unevenly-indented lines for readability. YAML's folding rules only fold line breaks between *equally*-indented lines - breaks next to a more-indented continuation line are preserved as real newlines. That meant the Jinja expression Ansible actually received still had real newlines embedded in its own text, immediately before the `join('\n')` call - and Ansible's templating turned that one `join` into a literal two-character `\n` in the rendered output instead of a real line break. This only showed up in real Ansible templating, not plain Jinja2 tested in isolation.

## Fix

- `| trim` on `github_keys` at the point of use, plus `| select` (no test) as a second, defensive layer against any other empty entry ending up in the list.
- Rewrote `permitted_keys_for_account` as a single line instead of a folded block scalar - the YAML-folding interaction above is why the multi-line version broke.

## Testing

Reproduced both bugs and the fix locally first (pure Jinja/Ansible templating logic, no AWS/live host needed), then Brenda re-ran the real `ansible-playbook --check --diff` against `openaustralia.org.au` and confirmed the fix.

### DIFF
--- a/roles/internal/deploy-user/tasks/main.yml
+++ b/roles/internal/deploy-user/tasks/main.yml
@@ -62,6 +62,10 @@
     exclusive: true
     key: "{{ permitted_keys_for_account }}"
   vars:
+    # This block's last content line ("{{ line }} {{ result.item }}") ends with a real
+    # newline of its own - trim_blocks only trims the newline right after a {% %} tag,
+    # not this one, so the block still comes out newline-terminated regardless of
+    # {%- / -%} on the surrounding for-tags. Trimmed at the point of use below instead.
     github_keys: |-
       {% for result in github_keys_fetched.results -%}
       {% for line in result.content.splitlines() if line | trim %}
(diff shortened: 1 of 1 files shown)


Br3nda:
I have already applied this change

benrfairless (approved):

PR #714, opened 2026-08-29
GitHub OIDC-federated IAM roles for deploying over SSM instead of direct SSH

Br3nda:
## Description

Prep for closing port 22 on the `webserver` security group: gives the openaustralia repo's
GitHub Actions deploy workflows a way to reach the server that doesn't depend on it, by reusing
the same `AWS-StartSSHSession` Session Manager mechanism `bin/ssh-config` already sets operators
up with (README "Accessing servers") - just driven by a scoped IAM role assumed via GitHub OIDC
instead of a person's own AWS profile. No long-lived AWS credentials need to live in GitHub as a
result.

Two roles (`github-actions-deploy-staging`/`-production`), each trusted only for its own GitHub
Environment's OIDC subject claim - mirrors why `openaustralia_deploy_ci_keys`
(`group_vars/openaustralia.yml`) is already two separate SSH keys rather than one shared between
stages: a compromised staging run shouldn't be able to assume the production role too. Each role
can only `ssm:StartSession` against the one openaustralia instance (both stages share it, per
`production.tf`'s own comment) via the `AWS-StartSSHSession` document specifically, plus
read-only `ec2:DescribeInstances` so the workflow can resolve the instance's current ID by its
`PublicHostname` tag at run time rather than hardcoding one that goes stale across a blue/green
cutover.

This only changes *how the network connection is made*. Once connected, everything else is
unchanged: the `deploy-user` role's `authorized_keys` and the `DEPLOY_SSH_KEY` GitHub Environment
secret still do the actual SSH authentication as the `deploy` user, and Capistrano's own SSH
config (`openaustralia/config/deploy.rb`) is untouched - Net::SSH reads `~/.ssh/config` by
default, so a `ProxyCommand` there is transparent to it.

Related to openaustralia/openaustralia#934 (the auto-deploy issue that added the SSH-based
deploy workflows this replaces the transport for). Companion PR on the workflow side:
openaustralia/openaustralia#935 (adds the matching
`configure-aws-credentials` + SSM `ProxyCommand` steps to `deploy-staging.yml`/
`deploy-production.yml`).

## Motivation and Context

We're closing port 22 on the webserver security group. The GitHub Actions deploy workflows
currently connect over a direct network SSH connection (`ssh-keyscan` + plain `ssh`/Capistrano),
which has no fallback once that port is gone - unlike operator access, which already migrated to
SSM. This is the AWS-side half of giving deploy the same migration path.

## How Has This Been Tested?

- [x] Ran automated tests on my own system - `terraform fmt -check`, `terraform validate`
      (both clean)
- [ ] Checked affected area manually on my own / staging system - **not yet applied**: no AWS
      credentials available where this was written, and every `tf-apply` is treated as a live
      production action per this repo's own AGENTS.md. Needs `make tf-plan` reviewed and
      `make tf-apply` run by someone with real credentials before the workflow-side PR can
      actually use these roles.

## Types of Changes

## Checklist


terraform/github-actions-deploy.tf
```diff
@@ -0,0 +1,121 @@
...
+
+  role   = aws_iam_role.github_actions_deploy[each.key].name
+  name   = "ssm-session-to-openaustralia-server"
+  policy = data.aws_iam_policy_document.github_actions_deploy.json
+}
+
+output "github_actions_deploy_role_arns" {
+  value       = { for k, r in aws_iam_role.github_actions_deploy : k => r.arn }
+  description = "Paste each into the matching openaustralia repo GitHub Environment as the AWS_DEPLOY_ROLE_ARN variable (Settings > Environments > staging/production > Variables - not a secret, an IAM role ARN isn't sensitive on its own)"
+}
```
Br3nda, line 121: we can manage the github secrets with terraform as well

terraform/github-actions-deploy.tf
```diff
@@ -0,0 +1,121 @@
...
+    actions = [
+      "ssm:StartSession",
+      "ssm:TerminateSession",
+      "ssm:ResumeSession",
+    ]
+    resources = [
+      module.openaustralia.production_instance_arn,
+      "arn:aws:ssm:${var.ec2_region}::document/AWS-StartSSHSession",
+    ]
+  }
```
ianheggie-oaf: `ssm:TerminateSession`/`ResumeSession` are checked against a `session/...` resource ARN, not the instance/document ARNs already listed here — as written, either action could be denied even though `StartSession` succeeds.

```suggestion
data "aws_caller_identity" "current" {}

data "aws_iam_policy_document" "github_actions_deploy" {
  statement {
    # Only this one instance, and only the SSH-port-forwarding session
    # document - not general Session Manager shell access (a different
    # document, AWS-StartInteractiveCommand / the default port-forwarding-only
    # AWS-StartSSHSession is already the narrowest of the standard documents
    # for this purpose).
    effect = "Allow"
    actions = [
      "ssm:StartSession",
      "ssm:TerminateSession",
      "ssm:ResumeSession",
    ]
    resources = [
      module.openaustralia.production_instance_arn,
      "arn:aws:ssm:${var.ec2_region}::document/AWS-StartSSHSession",
      "arn:aws:ssm:${var.ec2_region}:${data.aws_caller_identity.current.account_id}:session/$${aws:username}-*",
    ]
  }
```

ianheggie-oaf (commented):
A few more notes, for you to action manually since they can't be posted as GitHub suggestion blocks:

**Commit trailer** — not a file line, so no suggestion block, but per AGENTS.md this should read:
```
```
in place of the `🤖 Generated with...` / `Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>` lines.

**`terraform/righttoknow/dns.tf`** — isn't part of this PR's diff at all (it's an uncommitted local change on the branch: staging DNS `proxied` false→true), so there's nothing here to attach a suggestion to. Recommend reverting it on this branch and raising it as its own PR so it's reviewed separately:
```
  proxied = false
```
(both the `staging` and `www_staging` `cloudflare_record` blocks in that file)

**Port 22 closure** — agreed this should be its own PR, once all app repos are validated as deploying over SSM and their GitHub Actions workflows are updated. Worth softening the top-of-file comment here too, since as written it reads as if this PR alone is the prep step — righttoknow and theyvoteforyou still share `aws_security_group.webserver` and have no equivalent role yet.

**Thumbprint** — `thumbprint_list` hardcodes GitHub's OIDC root CA thumbprint with nothing elsewhere pointing at it if it goes stale. Low severity (AWS validates the chain itself for well-known providers), but worth knowing where to look if `AssumeRoleWithWebIdentity` ever starts failing.

**Shared module** — this OIDC-provider-plus-per-environment-role pattern is hand-rolled for openaustralia only. planningalerts/righttoknow/theyvoteforyou are provisioned through parallel modules and will likely want the same treatment eventually — worth a shared module rather than copy-pasting this file (and re-copying the session-ARN bug below) when that happens. Not blocking for this PR.

ianheggie-oaf (changes requested):
Various requested changes from claude and I

PR #716, opened 2026-08-29
Separate, non-secret Sentry key for browser-side tracking

Br3nda:
## Description

Companion to openaustralia/twfy#234 (browser-side Sentry error tracking). That PR initially
reused the existing server-side \`SENTRY_DSN\`'s public key directly - not a confidentiality
problem (a DSN's public key [isn't secret](https://docs.sentry.io/product/security-legal-pii/security-policy-reporting/#is-the-dsn-a-secret),
it's designed to sit in client-visible code) - but on reflection, an abuse-surface one: the exact
same key server-side error reporting uses would sit in every visitor's page source, and if
someone scraped it and spammed fake events, they'd land in the *same* project as server-side
monitoring, potentially burying real errors.

Adds \`sentry_browser_public_key\` - a second **Client Key on the same Sentry project**
(\`sentry_project_id\`), not a new project - so the browser key is independently revocable/
rate-limitable without touching server-side monitoring. \`sentry_browser_dsn\` builds the full DSN
from it (same shape as the existing \`sentry_dsn\`), and \`general.j2\` exposes it as a new
\`SENTRY_BROWSER_DSN\` PHP constant.

## Before this is fully live

\`sentry_browser_public_key\` is currently an empty string - I can't generate a Sentry Client Key
myself (needs the Sentry dashboard, not Terraform/Ansible). Someone with Sentry access needs to:

1. Sentry dashboard > Settings > Projects > (this project) > Client Keys > **Generate New Key**
   (don't reuse the existing one - that's the whole point).
2. Replace the empty string with that key's public part.

Empty is safe to merge/apply as-is in the meantime: \`general.j2\`'s own \`| default('')\` means an
empty \`sentry_browser_public_key\` just keeps the browser SDK disabled (\`SENTRY_BROWSER_DSN\`
stays empty), same behaviour \`SENTRY_DSN\` itself already has when unconfigured - openaustralia/twfy#234's
own guard (\`defined('SENTRY_BROWSER_DSN') && SENTRY_BROWSER_DSN\`) already expects this.

Deliberately **not** vault-encrypted like \`sentry_secret_key\` - unlike that one, there's no
confidentiality this value could protect, so vault-encrypting it would just add friction with no
security benefit.

## Motivation and Context

See openaustralia/twfy#234 and openaustralia/openaustralia... (conversation with Claude Code) for
the fuller back-and-forth on why a second key rather than reusing the existing one.

## How Has This Been Tested?

- [x] Ran automated tests on my own system - \`.venv/bin/yamllint\` on the two \`group_vars\` files
      and the template (no new warnings - reflowed one line as a YAML folded scalar to stay under
      the 120-char limit), and the Jinja conditional itself tested directly with both an empty and
      a populated \`sentry_browser_public_key\` (confirms empty renders as \`''\`, not a malformed
      \`https://@host/id\`)
- [ ] Checked affected area manually on my own / staging system - can't fully verify end-to-end
      until a real Client Key exists (see "Before this is fully live" above)

## Types of Changes

## Checklist


Br3nda:
No longer needed — openaustralia/twfy#234 was reverted to reuse the existing server-side SENTRY_DSN for the browser SDK instead of provisioning a separate key.

